PR #9, opened 2019-01-10
Assignment Submission - Raymond Shiner & Bryan Caeser

raymondshiner:


Sturminator:
Looking over your code, it seems we are in similar spots. We aren't passing the checks because we haven't fully implemented our DbContext. Other than that, I think your work looks solid so far.

SecretSanta/src/SecretSanta.Domain/Interface/iEntity.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Interface
+{
+    public interface IEntity
```
breaman: Filenames should match the name of the classes/interfaces.

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            if (theGift == null)
+                return;
+
+            User dbUser = DbContext.Users.Find(userId);
+
+            if (dbUser == null)
+                return;
+
+            theGift.User = dbUser;
```
breaman: If the Gift class had a property of UserId, then you would not have to fetch the user from the database before making the association. There is nothing wrong with this, but it does save one trip to the database. If a user with the UserId doesn't exist in the database, then an exception would be thrown due to foreign key constraints.

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public Gift FindGift(int giftId)
+        {
+            return DbContext.Gifts.Find(giftId);
+        }
+
+
+        //Create not required, we made it because we thought it might be useful later
+        /*
+        public Gift CreateGift(string title, int importance, string description, string url)
```
breaman: This will be a good teaching point for class on Thursday, but there are other ways to enforce these types of constraints without having to put it in the service method. We will talk about using attributes. Ideally, you really don't need a service method for just creating a gift because the Gift class should know how to create itself and have the necessary constraints there.

SecretSanta/test/SecretSanta.Domain.Tests/ServiceTests/MessageServiceTests.cs
```diff
@@ -0,0 +1,107 @@
...
+
+            using (var context = new ApplicationDbContext(Options))
+            {
+                context.Database.EnsureCreated();
+            }
+
+            CreateTestMessage();
+        }
+
+        private void CreateTestMessage()
```
breaman: This would be better to not create a property on the class and act more like the GiftService test class where you would return a Message object from this method instead.

raymondshiner: So why exactly is that better?

PR #10, opened 2019-01-10
Initial commit, following Stokes.

bcaesar42:


SecretSanta/src/SecretSanta.Domain/Models/Tag.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Models
+{
+    public class Tag
```
Cwhite534: Tag isn't required for this assignment. It was only used in Mikes lecture example.

SecretSanta/src/SecretSanta.Domain/Models/User.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Models
+{
+    public class User :Entity
+    {
+        public string FirstName { get; set; }
+        public string LastName { get; set; }
+        public ICollection<Post> Posts { get; set; }
```
Cwhite534: Similarly, Posts are also not required in this assignment.

PR #14, opened 2019-01-10
Assignment 1 - Christopher Sturm

Sturminator:
Test Merge


SecretSanta/src/SecretSanta.Domain/Models/User.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Models
+{
+    public class User : Entity
+    {
+        public string FirstName { get; set; }
+        public string LastName { get; set; }
+        public ICollection<Gift> Gifts { get; set; }
```
bcaesar42: You should be able to access the Gifts associated with a User by using the User's id to query the database. Therefore, it may be redundant for a User to have a direct reference to its gifts.

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,36 @@
...
+using SecretSanta.Domain.Models;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Services
+{
+    public class GiftService
+    {
+        private SecretSantaDbContext _context { get; set; }
```
bcaesar42: According to the naming convention for properties, the property's name should be pascal case. Prefixing a variable name with an underscore suggests that the variable is a field instead of a property.

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -0,0 +1,60 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace SecretSanta.Domain.Services
+{
+    public class GroupService
+    {
+        private SecretSantaDbContext _context { get; set; }
```
bcaesar42: Consider making the property's name pascal case.

Sturminator: Yeah, after reviewing the guidelines, I've fixed this. Thanks for the suggestion.

SecretSanta/src/SecretSanta.Domain/Models/UserGroup.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace SecretSanta.Domain.Models
 {
-    public class UserGroup 
+    public class UserGroup : Entity
```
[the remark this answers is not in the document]
Sturminator: I agree and understand that UserId and GroupId could be used as primary keys so I may remove the Entity class extension in the future, but I have decided to keep it for the time being at least.

SecretSanta/src/SecretSanta.Domain/Models/User.cs
Sturminator: I've removed the gift collection after reviewing the code. You're right, It is redundant for User to have a reference to it.

PR #15, opened 2019-01-10
Assignment 1 - Brad Howard

BHoward85:
There are a few helper functions to ensure testable code for the objects and services.


BHoward85:
Not all Unit Test are working at this moment, UserService and GiftService pass their tests.

SecretSanta/src/SecretSanta.Domain/Interfaces/IEntity.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Interfaces
+{
+    public interface IEntity
+    {
+        int Id { get; set; }
+        string EntityType { get; set; }
```
grantdwoods: I noticed you aren't really using the EntityType property anywhere in your code. Is there a benefit to using this property in place or normal type checking?

BHoward85: I set this up to help differentiate Id number should I have to make a join table of mixed types, this was when I was under the impression that we might do SQL calls and view the database. But now that I look at it, this property is a leftover from an earlier idea on the class side of the homework.

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,68 @@
...
+    public class GiftService
+    {
+        private SecretSantaDbContext DbContext { get; }
+
+        public GiftService(SecretSantaDbContext context)
+        {
+            DbContext = context;
+        }
+
+        public void CreateGift(User user, Gift gift)
```
breaman: This isn't really creating the gift, it is adding a gift object the database. Not marking down, but might want to be a little more clear on what the method is actually doing. AddGift would probably have been a better name.

breaman: The signature here seems a little overly complicated. Why not set the User on the Gift object before even passing it in and then you wouldn't have to add it to the collection later. What happens if User or Gifts on User is null? Also, looking at your unit test, you are already setting the User object on the Gift, which is going to then automatically add the Gift to the User.Gifts collection.

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        public void EditGift(User user, Gift gift)
+        {
+            if (gift.Id != default(int))
+                DbContext.Gifts.Update(gift);
+            else
+                DbContext.Gifts.Add(gift);
+
+            List<Gift> set = (List<Gift>)user.Gifts;
```
breaman: What is the purpose behind this logic? You have already updated the gift in the database. Did you try not doing this and you saw weird behavior?

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        public bool HasGift(User user, Gift gift)
+        {
+            return user.Gifts.Contains(gift);
+        }
+
+        public Gift FindGift(User user, int id)
+        {
+            return null;
```
breaman: What is this method here for? Seems like it isn't doing anything that would be intended.

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            user.Gifts = set;
+            DbContext.SaveChanges();
+        }
+
+        public void DeleteGift(User user, Gift gift)
+        {
+            DbContext.Gifts.Remove(gift);
+
+            user.Gifts.Remove(gift);
```
breaman: Not sure you want to be trying to do "sync" logic here. What happens if the SaveChanges call fails, you are going to have a gift that exists in the database but not on the user object. This whole service really should have been more of a single unit of responsibility and should not be manipulating the user at all.

SecretSanta/test/SecretSanta.Domain.Tests/GiftServiceTests.cs
```diff
@@ -0,0 +1,149 @@
...
+                giftService = new GiftService(context);
+
+                giftService.CreateGift(user, gift);
+            }
+
+            using (var context = new SecretSantaDbContext(Options))
+            {
+                giftService = new GiftService(context);
+
+                Assert.AreEqual<bool>(true, giftService.HasGift(user, gift));
```
breaman: I am not sure what this is actually testing. Looking at the HasGift method, it is just making sure that the gift exists in the user.Gifts collection, it isn't validating that the gift was actually persisted.

SecretSanta/test/SecretSanta.Domain.Tests/GiftServiceTests.cs
```diff
@@ -0,0 +1,149 @@
...
+                gift.URL = "amazon.com/Toaster/";
+
+                giftService.EditGift(user, gift);
+            }
+
+            using (var context = new SecretSantaDbContext(Options))
+            {
+                giftService = new GiftService(context);
+
+                if(giftService.HasGift(user, gift))
```
breaman: These tests should be fetching the object back out of the database to make sure everything was persisted correctly.

SecretSanta/src/SecretSanta.Domain/Services/MessageService.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class MessageService
+    {
+        private SecretSantaDbContext DbContext { get; }
+
+        public MessageService(SecretSantaDbContext context)
+        {
+            DbContext = context;
+        }
+
+        public void StoreMassage(Message message)
```
breaman: Careful of typo's. I believe this method should have been StoreMessage instead.

SecretSanta/test/SecretSanta.Domain.Tests/MessageServiceTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        [TestCleanup]
+        public void CloseConnection()
+        {
+            SqliteConnection.Close();
+        }
+
+        [TestMethod]
+        public void StoreMassage()
```
breaman: Looks like your test method had the same typo in the name. these can come back to haunt you.

SecretSanta/test/SecretSanta.Domain.Tests/MessageServiceTests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            Message message = new Message() { UserFrom = uFrom, UserTo = uTo, MessageBody = body };
+
+            using (var context = new SecretSantaDbContext(Options))
+            {
+                messageService = new MessageService(context);
+                userService = new UserService(context);
+
+                userService.UpsertUser(uFrom);
+                userService.UpsertUser(uTo);
```
breaman: You actually don't need these calls, if you just set the users on the message.UserTo and message.UserFrom, it will create the users for you if they don't exists and make the proper associations.

SecretSanta/test/SecretSanta.Domain.Tests/MessageServiceTests.cs
```diff
@@ -0,0 +1,70 @@
...
+                message.UserToId = uTo.Id;
+                message.UserFromId = uFrom.Id;
+
+                messageService.StoreMassage(message);
+            }
+
+            using (var context = new SecretSantaDbContext(Options))
+            {
+                messageService = new MessageService(context);
+                userService = new UserService(context);
```
breaman: userService isn't used in this scope anywhere.

SecretSanta/test/SecretSanta.Domain.Tests/PairingServiceTests.cs
```diff
@@ -0,0 +1,71 @@
...
+                pairingService = new PairingService(context);
+                userService = new UserService(context);
+
+                Pairing pair = new Pairing { Santa = santa, UserFor = reciver };
+
+                userService.UpsertUser(santa);
+                userService.UpsertUser(reciver);
+
+                pair.SantaId = santa.Id;
+                pair.UserForId = reciver.Id;
```
breaman: Since you are already setting Santa and UserFor in the creation of the Pairing object, you don't need the UpsertUser calls or the setting of the Id's

PR #18, opened 2019-01-14
Assignment1 - Eric Fueng

efueng:


SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Text;
+
+namespace SecretSanta.Domain.Models
+{
+    public class Gift : Entity
+    {
+        private string _Title;
+        public string Title
+        {
+            get => _Title;
```
KennethWhite: This is purely a style thing, but you can leave this line as auto implemented
```suggestion
            get;
```

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,49 @@
...
+            set
+            {
+                if (value is null)
+                {
+                    throw new ArgumentNullException("value");
+                }
+
+                value = value.Trim();
+
+                if (value is "")
```
KennethWhite: You can condense these two checks down using `string.IsNullOrEmpty(value)`

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,49 @@
...
+        private int _OrderOfImportance;
+        public int OrderOfImportance
+        {
+            get => _OrderOfImportance;
+            set
+            {
+                if (value < 1)
+                {
+                    throw new ArgumentOutOfRangeException("OrderOfImportance must be at least 1.", "value");
+                }
```
KennethWhite: It looks like you forgot to actually assign into the backing field here after your check

SecretSanta/src/SecretSanta.Domain/Models/Group.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            get => _Title;
+            set
+            {
+                if (value is null)
+                {
+                    throw new ArgumentNullException("value");
+                }
+
+                value = value.Trim();
```
KennethWhite: Again a stylistic critique, but this can be condensed down into the assignment below
```suggestion
                _Title = value.Trim();
```
And again you can turn the two separate checks into a single if statement

SecretSanta/src/SecretSanta.Domain/Models/Group.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public void AddUser(string firstName, string lastName)
+        {
+            User user = new User(firstName, lastName);
+            Users.Add(user);
+        }
+
+        public void RemoveUser(string firstName, string lastName)
+        {
+            foreach(User u in Users)
```
KennethWhite: You have written your own remove method here, rather than use the default implementation in `List`.

SecretSanta/src/SecretSanta.Domain/Models/User.cs
```diff
@@ -0,0 +1,61 @@
...
+            set
+            {
+                if (string.IsNullOrWhiteSpace(value))
+                {
+                    throw new ArgumentException("value");
+                }
+
+                value = value.Trim();
+
+                //if (value is "")
```
KennethWhite: I see here you did condense your checks into a single `if` clause, but left the commented code here. It should be removed by the final PR if it isn't used.

SecretSanta/src/SecretSanta.Domain/Models/User.cs
```diff
@@ -0,0 +1,61 @@
...
+        private string _FirstName;
+
+        public string LastName
+        {
+            get => _LastName;
+            set
+            {
+                if (value is null)
+                {
+                    throw new ArgumentNullException("value");
```
KennethWhite: I don't think the error message generated here is descriptive enough for debugging purposes, consider adding a description to this.

SecretSanta/test/SecretSanta.Domain.Tests/Models/UserTests.cs
```diff
@@ -0,0 +1,30 @@
...
+            Assert.AreEqual("Edmond", user.FirstName);
+            Assert.AreEqual("Dantes", user.LastName);
+        }
+
+        [TestMethod]
+        [DataRow(null)]
+        [DataRow(default(string))]
+        [DataRow(" ")]
+        [ExpectedException(typeof(ArgumentException))]
+        public void FirstName_AssignedNullOrEmpty_ThrowsArgumentException(string value)
```
KennethWhite: For the sake of thoroughness you should probably do this for the `LastName` as well.

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
efueng: I tried this and Visual Studio yelled at me. Maybe because there's a private backing field the setter uses?

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            DbContext.Gifts.Add(gift);
+            DbContext.SaveChanges();
+
+            return gift;
+        }
+
+        public void AddGiftForUser(int userId, Gift gift)
+        {
+            User user = DbContext.Users.Find(userId);
```
breaman: Since your Gift object has a UserId on it, you could have just done
```
gift.UserId = userId;
```

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,53 @@
...
+            user.Gifts.Add(gift);
+
+            DbContext.SaveChanges();
+        }
+
+        public void RemoveGiftForUser(int userId, Gift gift)
+        {
+            User user = DbContext.Users.Find(userId);
```
breaman: Removing the gift from the DB would also automatically remove it from the user.Gifts collection.

SecretSanta/test/SecretSanta.Domain.Tests/Services/GiftServiceTests.cs
```diff
@@ -0,0 +1,173 @@
...
+                var newGift = new Gift
+                {
+                    Title = "Another blaster",
+                    OrderOfImportance = 2,
+                    Description = "More toys for Wedge",
+                    Url = "www.morepewpew.com",
+                    User = gift.User
+                };
+
+                service.AddGiftForUser(gift.UserId, newGift);
```
breaman: Setting the User = gift.User on line 89, it already makes the association for the user and the gift, so this method is a bit of overkill and is already technically covered in your AddGift method.

SecretSanta/test/SecretSanta.Domain.Tests/Services/MessageServiceTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        [TestMethod]
+        public void FindMessage()
+        {
+            using (var context = new ApplicationDbContext(Options))
+            {
+                MessageService service = new MessageService(context);
+                var message = CreateMessage();
+                var persistedMessage = service.AddMessage(message);
+                var fetchedMessage = service.Find(1);
```
breaman: This is not doing what you think it is doing. We will cover in class on Thursday because it will be a good teaching point for everyone.

PR #20, opened 2019-01-16
Prototype

Chester-Southwood:


SecretSanta/src/Models/Gift.cs
```diff
@@ -0,0 +1,26 @@
...
+    public class Gift : Entity
+    {
+        public string Title { get; set; }
+        public int OrderOfImportance { get; set; }
+        public string Url { get; set; }
+        public string Description { get; set; }
+        public User User { get; set; }
+        public int UserId { get; set; }
+
+        /*public Gift(string title, int orderOfImportance, string url, string description, User user)
```
efueng: I'm guilty of this too, but commented out code is a code smell. If you don't need it, you should remove it before the final PR.

SecretSanta/src/Services/GiftsService.cs
```diff
@@ -0,0 +1,82 @@
+﻿using src.Models;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace src.Services
+{
+    class GiftsService
+    {
+        private ApplicationDbContext _db { get; }
```
efueng: Small style issue, but I think property names are supposed to be Pascal case in C#.

SecretSanta/src/Services/GroupsService.cs
```diff
@@ -0,0 +1,76 @@
...
+
+
+        private Group FindGroup(int groupId)
+        {
+            return Db.Groups.Find(groupId);
+        }
+
+        public void AddGroup(Group group)
+        {
+            //Group foundGroup = FindGroup(id);
```
efueng: Usual warning about commented out code and code smells

SecretSanta/src/Services/UsersService.cs
```diff
@@ -0,0 +1,123 @@
...
+    {
+        private ApplicationDbContext Db { get; }
+
+        public UsersService(ApplicationDbContext db)
+        {
+            Db = db;
+        }
+
+        //not required
+        /*public void RemoveUser(User user)
```
efueng: Usual warnings about commented out code and code smells.

SecretSanta/src/Services/UsersService.cs
```diff
@@ -0,0 +1,123 @@
...
+
+                Db.Remove.
+                var saveChanges = Db.SaveChangesAsync();
+                saveChanges.Wait();
+            }
+        }*/
+
+        public void UpdateUser(User user)
+        {
+            if (IsNameNotNull(user) || IsUserNull(user))
```
efueng: You can avoid the if/else by checking against the complement of your original conditional. Also, consider checking for a null user first. As is, you could encounter a case where you are checking the name of a null user; changing the order would also short-circuit the evaluation in those cases.

```
if (user not null && user name not null)
{
    do stuff...
}
```

SecretSanta/src/Models/User.cs
```diff
@@ -0,0 +1,17 @@
+﻿using src;
+using System;
+using System.Collections.ObjectModel;
+
+namespace src.Models
+{
+    public class User : Entity
+    {
+        public string FirstName { get; set; }
```
efueng: I think this comes down to implementation styles/preferences, but if you validate input in your properties here you can avoid having to write null checking methods such as those in your UserService.

efueng:
Looks pretty good to me Chester. The only issue is an easily-fixed code smell (besides the build pipeline tests). I pointed out some spots where you might consider alternative implementations, but overall it looks like you're almost done!

SecretSanta/src/Model/Entity.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace src.Model
+{
+    public class Entity
+    {
+        public int Id { get; set; }
+        public static void Main()
```
breaman: There is no need for a main method here, this class is not supposed to be an executable.

SecretSanta/src/Services/GiftService.cs
```diff
@@ -0,0 +1,97 @@
...
+        public Gift Find(int giftId)
+        {
+            return Db.Gifts
+                .Include(g => g.User)
+                .SingleOrDefault(g => g.Id == giftId);
+        }
+
+        public bool Add(Gift gift)
+        {
+            Gift possibleGift = Find(gift.Id);
```
breaman: What is this line supposed to be doing, checking to see if the Id already exists in the database? EF will take care of that for you by throwing an exception, but if you really wanted to make this check, just verify that gift.Id is 0. It will save a roundtrip database call.

SecretSanta/src/Services/GiftService.cs
```diff
@@ -0,0 +1,97 @@
...
+            return possibleGift;
+        }
+
+        /*
+        public bool IsUserNull(User user)
+        {
+            return user == null;
+        }*/
+
+        public bool IsGiftNull(Gift group)
```
breaman: Why have this method? Also, be careful of naming, your parameter name here is group, but you are passing a Gift. That can be confusing for people using your library.

SecretSanta/src/Services/GiftService.cs
```diff
@@ -0,0 +1,97 @@
...
+        }*/
+
+        public Gift Delete(Gift possibleGift)
+        {
+            if (possibleGift != null)
+            {
+                Db.Gifts.Remove(possibleGift);
+                Db.SaveChangesAsync().Wait();
+            }
+            return possibleGift;
```
breaman: Since you are deleting this gift, I'm not sure you really want to be returning it. What would be the purpose for passing back a gift that shouldn't exist anymore?

SecretSanta/src/Services/GiftService.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            User foundUser = Db.Users.Find(userId);
+            foundUser.GiftList.Remove(gift);
+            var saveChanges = Db.SaveChangesAsync();
+            saveChanges.Wait();
+        }*/
+
+        public Gift Delete(Gift possibleGift)
+        {
+            if (possibleGift != null)
```
breaman: Noticed you didn't use the IsGiftNull method here. I would have suggested this is how the code should have looked the whole time and the IsGiftNull method should not exist.

SecretSanta/src/Services/GiftService.cs
breaman: Is this method supposed to ever be called outside of this class? If not, I would suggest that you make it private instead of public.

SecretSanta/src/Model/ApplicationContext.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.EntityFrameworkCore;
+using src.Model;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace src.Models
```
breaman: Why have the Model and Models namespace? These should have all been in the Models namespace to begin with.

SecretSanta/src/Services/MessageService.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        public Message Find(int id)
+        {
+            return Db.Messages
+                .Include(message => message.Pairing)
+                .SingleOrDefault(message => message.Id == id);
+        }
+
+        public void StoreMessage(Pairing pair, Message message)
```
breaman: How is this different from Add? Wouldn't the message object have the associated pairing with it already, or are you allowing for a message to be saved off that doesn't have a sender and recipient?

SecretSanta/src/Services/MessageService.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            Db = db;
+        }
+
+        public bool Add(Message message)
+        {
+            if (!IsMessageNull(message))
+            {
+                Db.Messages.AddAsync(message).Wait();
+                Db.SaveChangesAsync();
```
breaman: Careful of using async versions of the method here. You aren't using a Wait() like you are on the line before so this might not behave quite the way you expect it to.

SecretSanta/src/Services/PairingService.cs
```diff
@@ -0,0 +1,56 @@
...
+        public PairingService(ApplicationDbContext db)
+        {
+            Db = db;
+        }
+
+        public bool Add(Pairing pairing)
+        {
+            if (!IsPairingNull(pairing))
+            {
+                Db.Pairs.AddAsync(pairing);
```
breaman: you are mixing async and non-async methods here. For our example this might not be an issue, but you really don't want to do this, it can cause some nasty behaviors.

SecretSanta/test/TestServices/TestGiftService.cs
```diff
@@ -0,0 +1,207 @@
...
+using System.Text;
+
+namespace test.TestServices
+{
+    [TestClass]
+    public class TestGiftsService
+    {
+        private SqliteConnection SqliteConnection { get; set; }
+        private DbContextOptions<ApplicationDbContext> Options { get; set; }
+        private GiftsService GiftService { get; set; }
```
breaman: There is no need for GiftService or Gift to be properties on the test class. I see you used the Gift in the [TestInitialize], but then you also recreated a new instance all the time in the test methods as well. Both GiftService and Gift should just live locally in each method.

SecretSanta/test/TestServices/TestMessageSystem.cs
```diff
@@ -0,0 +1,181 @@
...
+            User recep = new User
+            {
+                FirstName = "Gary",
+                LastName = "Oak",
+            };
+
+            Pairing pairing = new Pairing
+            {
+                Santa = santa,
+                SantaId = santa.Id,
```
breaman: here you would not want to set SantaId and RecipientId since those values are going to be 0 (or values before the item is in the database).

SecretSanta/src/Model/Pairing.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace src.Model
+{
+    public class Pairing : Entity
+    {
+        public User Santa { get; set; }
+        public int SantaId { get; set; }
+        public User Recepiant { get; set; }
```
breaman: Careful of typos, this is actually Recipient

SecretSanta/test/TestServices/TestMessageSystem.cs
```diff
@@ -0,0 +1,181 @@
...
+                Recepiant = recep,
+                RecepiantId = recep.Id,
+                Messages = new List<Message>()
+            };
+
+            Message message = new Message
+            {
+                MessagePost = "Heated Rivaly!",
+                Pairing = pairing,
+                PairingId = pairing.Id
```
breaman: Don't want to set the PairingId here either.

PR #22, opened 2019-01-16
Initial commit

RichardTeller:
Created Models, added AddUser service, added AddUser service test


SecretSanta/src/SecretSanta.Domain/Models/ApplicationDbContext.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    public class ApplicationDbContext : DbContext
+    {
+        public DbSet<User> Users { get; set; }
+        public DbSet<Gift> Gifts { get; set; }
+
+        public ApplicationDbContext(DbContextOptions<ApplicationDbContext> options) : base(options) { }
+
+        /*
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
```
KennethWhite: It looks like this is a placeholder from the blog example Mike made in class, since our assignment doesn't have `PostTags`. You should probably remove this.

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,62 @@
...
+
+                return true;
+            }
+            else
+            {
+                return false;
+            }
+        }
+
+        public bool DeleteGift(User user)
```
KennethWhite: I know this is commented out, but the logic is a little weird. It is a non-static method requiring an instance of `GiftService` so it won't have access to the gift in question. Using `this` should refer to the GiftService itself, not the gift.

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            if (user != null)
+            {
+                if (user.Gifts.Contains(this))
+                {
+                    user.Gifts.Remove(this);
```
KennethWhite: The `List.Remove()` function will return a bool on success/failure, so you can reduce this down to probably a single line.
```suggestion
                    return user.Gifts.Remove(GiftToRemove);
```

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Models
+{
+    public class Gift : Entity
+    {
+        public string Title { get; set; }
+        public int Importance { get; set; }
+        public string URL { get; set; }
```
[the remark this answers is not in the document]
RichardTeller: thanks for that!

SecretSanta/src/SecretSanta.Domain/Models/ApplicationDbContext.cs
RichardTeller: Yea it was there just to learn from for similarity.  Probably should have left that out of my PR.  Anyway, thanks for the heads up.

PR #23, opened 2019-01-16
Assignment1

COsborn2:


SecretSanta/src/SecretSanta.Domain/Services/UserService.cs
```diff
@@ -0,0 +1,42 @@
...
+    public class UserService
+    {
+        public UserService(ApplicationDbContext context)
+        {
+            DbContext = context;
+        }
+
+        private ApplicationDbContext DbContext { get; }
+
+        public User AddUser(User user)
```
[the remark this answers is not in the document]
COsborn2: You're right. I changed it from Upsert to Add, but Upsert is probably a better descriptor of what's going on.

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,43 @@
...
+        public Gift UpsertGift(Gift gift)
+        {
+            if (gift.Id == default(int))
+                DbContext.Gifts.Add(gift);
+            else
+                DbContext.Gifts.Update(gift);
+            DbContext.SaveChanges();
+            
+            return gift;
+        }
```
Isaac-Bliss: I never thought to do it this way, it makes a lot of sense

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -0,0 +1,54 @@
...
+        public GiftService(ApplicationDbContext context)
+        {
+            DbContext = context;
+        }
+
+        private ApplicationDbContext DbContext { get; }
+
+        public Gift UpsertGift(Gift gift)
+        {
+            if (gift.Id == default(int))
```
breaman: I know it may make things a little more verbose, but best practices are actually to always use {} around statements even if they are only one liners.

SecretSanta/test/SecretSanta.Domain.Tests/SecretSanta.Domain.Tests.csproj
```diff
@@ -0,0 +1,28 @@
...
+    <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    </PropertyGroup>
+
+    <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="2.2.1" />
```
breaman: technically you don't need this reference since SecretSanta.Domain references it and it will get pulled in automatically.

PR #24, opened 2019-01-16
Assignment1

wpcwilliams:
Still have a lot to do here...


SecretSanta/src/SecretSanta.Domain/Services/GiftServices.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        private void EditGift(User theUser)
+        {
+
+        }
+
+        private void DeleteGift(User theUser)
+        {
+
+        }
```
raymondshiner: So the purpose of these methods is to call to the DB and pull out a given user, and modify their gift somehow (add a gift, edit a gift, remove a gift). But you have the user passed in on all of these methods, which means you are currently expecting the call to the DB to made outside this method. I would refactor these to take in a userID maybe, or some other key that you can then use a find() on to pull them from the db.

SecretSanta/src/SecretSanta.Domain/Models/Message.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+
+namespace SecretSanta.Domain.Models
+{
+    public class Message : Entity
+    {
+        public User Sender { set; get; }
+        public User Reciever { set; get; }
```
breaman: Be careful of typos here, this property should be Receiver.

SecretSanta/test/SecretSanta.Domain.Tests/Services/ApplicationServiceTest.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+
+        [TestInitialize]
+        public void OpenConnection()
+        {
+            SqliteConnection = new SqliteConnection("DataSource=:memory:");
+            SqliteConnection.Open();
+
+            if (DEBUG)
```
breaman: I like the idea here, but if you wanted to do this in a real app, you would probably want to look into compiler directives. That way you don't have to change code to change behavior. https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if

PR #25, opened 2019-01-16
Assignment1

grantdwoods:


grantdwoods:
If anyone understands the GroupService well, I would appreciate feedback on how I am trying to add/remove a user from a group! Will write unit tests for this later today.

SecretSanta/test/SecretSanta.Domain.Tests/Services/UserServiceTest.cs
```diff
@@ -0,0 +1,105 @@
...
+        public void CloseConnection()
+        {
+            SqliteConnection.Close();
+        }
+        [TestMethod]
+        public void AddUser()
+        {
+            UserService userService;
+
+            User user = new User
```
ConnerV42: You could add a CreateUser class, in order to reduce code duplication in this class, similar to your Gift class'es CreateGift helper method.

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public Group AddGroup(Group group)
+        {
+            Context.Groups.Add(group);
+            Context.SaveChanges();
+
+            return group;
+        }
+
+        public void AddUserToGroup(int groupId, UserGroup userGroup)
```
ConnerV42: This looks like a good implementation. My only suggestion would be to possibly increase the encapsulation of this method by taking in the Entire Group object and then just grabbing the group Id from within the method.

SecretSanta/src/SecretSanta.Domain/Interface/IEntity.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Interface
```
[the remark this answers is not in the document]
grantdwoods: Good point, I will consider if there is a compelling reason to keep this structure around.

SecretSanta/test/SecretSanta.Domain.Tests/Services/UserServiceTest.cs
grantdwoods: That's a good point, I also forgot about C#'s default parameter option until I started looking at everyone's code.  I will definitely refactor this!

grantdwoods:
Having a hard time being able to filter out Users based on their collection of UserGroups. The last test in the GroupServiceTest is "passing" because there is nothing verifying the method works for now...

PR #34, opened 2019-01-22
Assignment2 - Eric Fueng

efueng:


SecretSanta/src/SecretSanta.Library/FileImport.cs
```diff
@@ -0,0 +1,62 @@
...
+                return Path.Combine(System.Reflection.Assembly.GetExecutingAssembly().Location, @"..\..\..\..\data\");
+            }
+        }
+            
+
+        public FileStream OpenFile(string path)
+        {
+            if (string.IsNullOrEmpty(path))
+            {
+                FileStream f = new FileStream(path, FileMode.Open);
```
Keboo: This will throw an exception if you pass a null or empty string to [FileStream's ctor](https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream.-ctor?view=netcore-2.2#System_IO_FileStream__ctor_System_String_System_IO_FileMode_)

SecretSanta/src/SecretSanta.Library/FileImport.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            var _path = Path.GetFullPath(Path.Combine(RelativePath, path));
```
Keboo: For local variable don't use any prefix. 
```suggestion
            var path = Path.GetFullPath(Path.Combine(RelativePath, path));
```

SecretSanta/src/SecretSanta.Library/FileImport.cs
```diff
@@ -0,0 +1,62 @@
...
+            return fileStream;
+        }
+
+        public string ReadFile(string path, out List<string> fileContents)
+        {
+            string line = string.Empty;
+            using (FileStream fileStream = OpenFile(path))
+            {
+                fileContents = new List<string>();
+                using (StreamReader streamReader = new StreamReader(fileStream))
```
Keboo: You could save yourself a little work by simply using the [StreamReader ctor](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=netcore-2.2#System_IO_StreamReader__ctor_System_String_) that takes in a file path as a string.

SecretSanta/src/SecretSanta.Library/FileImport.cs
```diff
@@ -0,0 +1,62 @@
...
+                {
+                    while (streamReader.Peek() >= 0)
+                    {
+                        fileContents.Add(streamReader.ReadLine());
+                    }
+                    line = streamReader.ReadLine();
+                }
+            }
+
+            return line;
```
Keboo: This return value does not appear to get used.

SecretSanta/src/SecretSanta.Library/FileImport.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        public string ReadFile(string path, out List<string> fileContents)
+        {
+            string line = string.Empty;
+            using (FileStream fileStream = OpenFile(path))
+            {
+                fileContents = new List<string>();
+                using (StreamReader streamReader = new StreamReader(fileStream))
+                {
+                    while (streamReader.Peek() >= 0)
```
Keboo: You might consider using [StreamReader.EndOfStream](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.endofstream?view=netcore-2.2)

SecretSanta/src/SecretSanta.Library/FileImport.cs
efueng: Yeah it's a very janky set up right now. I'm expecting an exception because in one of my tests I have data rows that pass a null and then an empty string. This will be refactored before the final PR deadline I promise!

SecretSanta/src/SecretSanta.Library/FileImport.cs
efueng: Would you mind elaborating as to why? Honestly curious because I found the Peek example in MS's documentation for .NET Core 2.2.

Also, given your advice to use the StreamReader ctor that takes a file path as a string, can I assume that under the hood the file path is then converted to a stream of some type?

SecretSanta/src/SecretSanta.Library/FileImport.cs
efueng: I'll clean that right up. Thanks for the comments!

SecretSanta/src/SecretSanta.Library/FileImport.cs
Keboo: So first of all, yes that ctor does [convert to a Stream under the hood](https://source.dot.net/#System.Private.CoreLib/shared/System/IO/StreamReader.cs,150). The advantage is that the StreamReader will be responsible for managing the stream rather than you. Typically I favor writing as little code as possible to get the job done. 

As for `Peek()` vs `EndOfStream`. If you [look at the code for each](https://source.dot.net/#System.Private.CoreLib/shared/System/IO/StreamReader.cs,150) you will notice that they are basically doing the exact same thing.  The difference is `Peek()` returns the character, while `EndOfStream` skips the retrieval of the character in the success case (ie. when you are not at the end of the stream). The bigger reason I favor `EndOfStream` is I find it easier to read. The property name is explicit about the condition I am checking for. While `Peek()` requires me to remember that it returns a `char` upcast to an `int` or `-1` when it has reached the end of stream. The way you have it written is perfectly functional, and written correctly. It is merely a code style thing to help it be more readable. Does that make sense?

efueng: Yes that makes complete sense, thanks for taking the time to explain that!

SecretSanta/src/SecretSanta.Library/FileImport.cs
efueng: This has been fixed

SecretSanta/src/SecretSanta.Library/FileImport.cs
efueng: fixed

SecretSanta/src/SecretSanta.Library/FileImport.cs
```diff
@@ -0,0 +1,71 @@
...
+                throw new ArgumentException("Path is null or empty.");
+            }
+
+            fileContents = new List<string>();
+
+            using (StreamReader streamReader = new StreamReader(path))
+            {
+                while (!streamReader.EndOfStream)
+                {
+                    fileContents.Add(streamReader.ReadLine());
```
Chester-Southwood: Not sure if I'm correct, but aren't we supposed to omit empty lines?

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,81 @@
...
+            }
+        }
+
+        private string _Url;
+        public string Url
+        {
+            get => _Url;
+            set
+            {
+                if (string.IsNullOrEmpty(value))
```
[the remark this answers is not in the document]
efueng: Yeah I always have an internal debate about this. In the end laziness usually wins lol.

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,81 @@
...
+                {
+                    throw new ArgumentException("URL cannot be null or empty.", nameof(value));
+                }
+
+                value = value.Trim();
```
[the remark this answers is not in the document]
efueng: good catch thanks Chester

SecretSanta/test/SecretSanta.Library.Tests/FileImportTests.cs
```diff
@@ -0,0 +1,142 @@
...
+{
+    [TestClass]
+    public class FileImportTests
+    {
+
+        private FileImport FileImporter { get; set; }
+        private static string TempDirectory { get; } = Path.GetTempPath();
+
+        [TestInitialize]
+        public void TestInitialize()
```
MarkMichaelis: Isn't each file used in a specific test?  Why create all the files in `TestInitialize` rather than just in the specific test that used it?

SecretSanta/src/SecretSanta.Library/FileImport.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            if (!match.Success)
+            {
+                throw new FormatException("Invalid header format. Headers must follow the format of 'Name: <FirstName> <LastName>'" +
+                    "or 'Name: <LastName>, <FirstName>'.");
+            }
+
+            string[] tokenizedNames;
+
+            User user = new User();
```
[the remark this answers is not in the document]
efueng: Oh dear... hindsight is 20/20 right?

SecretSanta/src/SecretSanta.Library/FileImport.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            return user;
+        }
+
+        public User ImportUserAndGifts(string path)
+        {
+            User user;
+            List<string> fileContents;
+
+            ReadFile(out fileContents);
```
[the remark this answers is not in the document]
efueng: This was mostly me experimenting. I just read the part about out params and wanted to try them.

SecretSanta/test/SecretSanta.Library.Tests/FileImportTests.cs
efueng: I think this was mostly due to still getting a handle on unit testing. In the later assignments I tend to avoid such monolithic test initializers (except for the PairingService tests)

SecretSanta/src/SecretSanta.Library/FileImport.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        public FileImport() { }
+
+        public FileImport(string filePath)
+        {
+            FilePath = filePath;
+        }
+
+        public void ReadFile(out List<string> fileContents)
```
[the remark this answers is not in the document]
efueng: Usually I would have left these methods private since I really just use them to build the method group ImportUserAndGifts, but I left them public for testing reasons. Is it acceptable to test private helper methods like this indirectly via thorough testing of the public method that uses the private methods?

PR #35, opened 2019-01-22
Assignment2 - Raymond Shiner

raymondshiner:
File is defined as being formatted correctly if Header Line is formatted correctly. As of right now This library is assuming that every line in the file after the first line will contain just the title of each gift to add, So it doesn't care about the format of file past the first line.

Did this one on my own.


SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,123 @@
...
+
+        /// <summary>
+        /// File is defined as being formatted correctly if Header Line is formatted correctly. As of right now
+        /// This library is assuming that every line in the file after the first line will contain just the title of each gift to add,
+        /// So it doesn't care about the format of header past the first.
+        /// </summary>
+        /// <param name="header"></param>
+        /// <returns></returns>
+
+        public bool FileHeaderIsFormattedCorrectly(string header)
```
Keboo: There is a programming principle called [Tell Don't Ask](https://deviq.com/tell-dont-ask/). By having a method like this you are effectively asking "is this string in the correct format?". This then causes you to repeat the functionality of breaking apart the string both in this method and the one above. This then means that if there were a change to the requirements of the header you would have to change the logic in two places. 

Better is to have this method simply split apart the string into its name part and return that or some invalid value (you will note that this will largely obsolete this method and leave you with just `ExtractNamesFromHeader`).

Even better than this is to implement something like the [TryParse methods](https://docs.microsoft.com/en-us/dotnet/api/system.int32.tryparse?view=netframework-4.7.2) that both return a boolean for success/failure and also return the value already processed.

SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        /// <summary>
+        /// Takes in the correctly formatted header file and returns tuple (firstname, lastname) of header file. If header formatted
+        /// incorrectly it will return tuple (null, null).
+        /// </summary>
+        /// <param name="header">Correctly formatted header file - 'Name: First Last' or 'Name: Last, First'</param>
+        /// <returns></returns>
+
+        public (string firstname, string lastname) ExtractNamesFromHeader(string header)
```
Keboo: Is this method public simply so you can unit test it? If so, I would suggest making it private and only indirectly unit it test it as you test `ReadInUser`

SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,123 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.EntityFrameworkCore.Migrations.Operations;
+using SecretSanta.Domain.Models;
+using SecretSanta.Domain.Services;
+
+namespace SecretSanta.Import
+{
+    public class FileImporter
```
Keboo: This class does not appear to need any instance data. Is there a reason to leave it as an instance class rather than make everything static?

SecretSanta/SecretSanta.Import.Tests/FileImporterTests.cs
```diff
@@ -0,0 +1,302 @@
...
+            Initialize_ValidTestFile2_MarkFNF();
+            Initialize_ValidTestFile3_MikeLNF();
+            Initialize_InvalidTestFile1_GarbageHeader();
+            Importer = new FileImporter();
+        }
+
+        private void Initialize_ValidTestFile1_RayFNF()
+        {
+            string path = System.Environment.CurrentDirectory;
+            path += @"\ValidFile1_RayFNF.txt";
```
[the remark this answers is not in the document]
Keboo: There are several reasons (in no particular order):
 - Path.Combine is cross platform and will handle the subtle change in path separator characters (\ on windows and / on *nix systems)
- Path.Combine will intelligently put in the path separators. Examine the difference between `@"first\" + @"\second"` and `Path.Combine(@"first\", @"\second");` (also note you will get different behavior on Window and *nix systems).
- Path.Combine handles rooted paths. Examine the result of `Path.Combine("foo", @"C:\Bar")` vs `"foo" + @"C:\Foo";

You can take look at the source code for it [here](https://source.dot.net/#System.Private.CoreLib/shared/System/IO/Path.cs,2d7263f86a526264). All of the methods are hyperlinked so you can click around.

SecretSanta/SecretSanta.Import.Tests/FileImporterTests.cs
```diff
@@ -0,0 +1,302 @@
...
+            var header = "Name: Raymond Shiner";
+
+            (string firstName, string lastName) = Importer.ExtractNamesFromHeader(header);
+
+            Assert.AreEqual("Raymond", firstName);
+            Assert.AreEqual("Shiner", lastName);
+        }
+
+        [TestMethod]
+        public void ExtractNamesFromHeader_ValidHeaderFirstNameFirstMike_CorrectOutput()
```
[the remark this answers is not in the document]
Keboo: That is fine, but you only need a single unit test per condition that you are testing. If you want to test the same condition with different inputs it is better to do a single data driven test rather than a bunch of individual test methods.

SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        /// <summary>
+        /// Reads in User from given filename and returns a tuple of a User Object with FirstName and LastName
+        /// Stated in header file and a list of gifts populated with titles and importance. Returns (null, null) if file not found or file
+        /// incorrectly formatted.
+        /// </summary>
+        /// <param name="filename"></param>
+        /// <returns></returns>
+
+        public static (User user, List<Gift> gifts) ReadInUser(string filename)
```
MarkMichaelis: Why not just return a `User` or `List<Gifts>` given that one points to the other?

SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            string[] lines;
+
+            try
+            {
+                var path = Path.Combine(System.IO.Path.GetTempPath(), filename);
```
MarkMichaelis: Why do you fully qualify Path and also not qualify Path in the same statment?

SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,131 @@
...
+                    lines = System.IO.File.ReadAllLines(path);
+                }
+
+                else // account for absolute path
+                {
+                    lines = System.IO.File.ReadAllLines(filename);
+                }
+            }
+
+            catch (Exception)
```
MarkMichaelis: You should almost never catch and `Exception`.

SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        public static (User user, List<Gift> gifts) ReadInUser(string filename)
+        {
+            string[] lines;
+
+            try
+            {
+                var path = Path.Combine(System.IO.Path.GetTempPath(), filename);
+
+                if (File.Exists(path)) // account for relative path
```
MarkMichaelis: I think this could be cleaner.  Check if `filename` exists and then appent `GetTempPath()` if it doesn't.  Then `ReadAllLines()`. 
Why do you check if `path` exists but not check if `filename` exists even though you do `ReadAllLines()` for both?  I don't think you need to for eather because `ReadAllLines()` checks for you.

SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,131 @@
...
+
+                else // account for absolute path
+                {
+                    lines = System.IO.File.ReadAllLines(filename);
+                }
+            }
+
+            catch (Exception)
+            {
+                return (null, null);
```
MarkMichaelis: It isn't clear why null user and null gift list is ever valid?  Why not throw an excpetion?

SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        /// <summary>
+        /// Takes in the correctly formatted header file and returns tuple (firstname, lastname) of header file. If header formatted
+        /// incorrectly it will return tuple (null, null).
+        /// </summary>
+        /// <param name="header">Correctly formatted header file - 'Name: First Last' or 'Name: Last, First'</param>
+        /// <returns></returns>
+
+        public static (string firstname, string lastname) ExtractNamesFromHeader(string header)
```
MarkMichaelis: I am glad to see you are considering tuples, but rarely, if ever, are they exposed in public methods.

SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,131 @@
...
+                return (null, null);
+            }
+
+            string firstName;
+            string lastName;
+
+            var fullName = header.Substring(6); //Cutting off "Name: "
+
+
+            var names = fullName.Split(" ");
```
MarkMichaelis: Does this work if there are two spaces in the name?

SecretSanta/SecretSanta.Import/FileImporter.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <summary>
+        /// File is defined as being formatted correctly if Header Line is formatted correctly. As of right now
+        /// This library is assuming that every line in the file after the first line will contain just the title of each gift to add,
+        /// So it doesn't care about the format of header past the first.
+        /// </summary>
+        /// <param name="header"></param>
+        /// <returns></returns>
+
+        public static bool FileHeaderIsFormattedCorrectly(string header)
```
MarkMichaelis: Multiple returns interpersed between if statements is hard to read and generally can be simplified.

SecretSanta/SecretSanta.Import.Tests/SecretSanta.Import.Tests.csproj
```diff
@@ -0,0 +1,29 @@
...
+    <WarningsAsErrors />
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702;</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.10.0" />
```
MarkMichaelis: Great job adding the analyzers!!!

SecretSanta/SecretSanta.Import.Tests/GlobalSuppressions.cs
```diff
@@ -0,0 +1,8 @@
+﻿
+// This file is used by Code Analysis to maintain SuppressMessage 
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given 
+// a specific target and scoped to a namespace, type, member, etc.
+
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "<Pending>")]
```
MarkMichaelis: Great job with having the GlovalSuppression.cs file.

SecretSanta/SecretSanta.Import/GlobalSuppressions.cs
```diff
@@ -0,0 +1,8 @@
+﻿
+// This file is used by Code Analysis to maintain SuppressMessage 
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given 
+// a specific target and scoped to a namespace, type, member, etc.
+
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Globalization", "CA1307:Specify StringComparison", Justification = "<Pending>")]
```
MarkMichaelis: You will want to always specify a justification or remove the rule entirely from the code analysis.

SecretSanta/SecretSanta.Import/SecretSanta.Import.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <ApplicationIcon />
+    <Win32Resource />
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
MarkMichaelis: Great to see this!!

PR #36, opened 2019-01-22
Assignment2

Chester-Southwood:
Prototype of Assignment2 Submission


SecretSanta/Library/Library.cs
```diff
@@ -0,0 +1,199 @@
...
+using System.Collections.Generic;
+using System.IO;
+
+namespace Library
+{
+    public class LibraryClass
+    {
+        static void Main(string[] args)
+        {
+
```
ConnerV42: I would take the main out of the class, just because it is not needed in this scenerio.

SecretSanta/LibraryTests/LibraryTests.cs
```diff
@@ -0,0 +1,131 @@
...
+            FileInfo fileInfo = new FileInfo(TempFileName);
+            fileInfo.Attributes = FileAttributes.Temporary;
+
+            StreamWriter reader = File.AppendText(TempFileName);
+            reader.WriteLine("Carrie White");
+            reader.Flush();
+            reader.Close();
+        }*/
+
+        [DataRow(@"C:\Users\User\Desktop\EWU-CSCD496-2019-Winter-Assignment1 (3)\EWU-CSCD496-2019-Winter-Assignment1\SecretSanta\LibraryTests\FileAb.txt")]
```
KyleBurgi: I'm not sure how you would go by it but using absolute paths on your machine isn't testable in other environments.

SecretSanta/Library/Library.cs
```diff
@@ -0,0 +1,199 @@
+﻿using src.Model;
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Library
+{
+    public class LibraryClass
```
ConnerV42: If you are going to go the static library route, you should make the class static as well.

SecretSanta/Library/Library.cs
```diff
@@ -0,0 +1,199 @@
...
+            IfCustomNullException(str, "String cannot be null when attempting to convert to trimmed array.");
+            string[] stringArray = str.Split(" ");
+            for(int index = 0; index < stringArray.Length; index++)
+            {
+                stringArray[index] = stringArray[index].Trim();
+            }
+            return stringArray;
+        }
+
+        public static User CreateUserFirstLast(in string firstLastLine)
```
ConnerV42: I bet you could combine your CreateUserFirstLast and CreateUserLastFirst methods into one dynamic method to handle both scenerios.

SecretSanta/LibraryTests/LibraryTests.cs
```diff
@@ -0,0 +1,131 @@
...
+using Library;
+using System.IO;
+using System;
+
+namespace LibraryTests
+{
+    [TestClass]
+    public class LibraryTests
+    {
+        public string TempFileName { get; set; }
```
ConnerV42: Looks like you never use either of the class level properties in your LibraryTests class, so you may not need them.

SecretSanta/LibraryTests/LibraryTests.cs
```diff
@@ -0,0 +1,131 @@
...
+        public string TempFileLocation { get; set; }
+
+        /*[TestInitialize]
```
ConnerV42: If you end up continuing along the route of generating a random file in the TestInitialize method, make sure that you include a TestCleanup method as well to take care of the file after the tests finish.

SecretSanta/Library/Library.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        public static void IfCustomNullException(in string path, in string customMessage)
+        {
+            if (path is null)
+            {
+                throw new ArgumentNullException(customMessage);
+            }
+        }
+
+        public static bool IsStringNull(in string str)
```
bcaesar42: I don't see the need for this method.
`if (str is null)` and `if (IsStringNull(str))` are about the same length to call, and there isn't any additional logic inside of the method.

PR #38, opened 2019-01-22
Assignment2 - Christopher Sturm

Sturminator:


SecretSanta/SecretSanta.Import/Services/GiftImportService.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            string path = Path.Combine(System.IO.Path.GetTempPath() + @"\", filename);
+
+            if (!File.Exists(filename))
+            {
+                if (!File.Exists(path))
+                {
+                    throw new FileNotFoundException("Could not find file " + filename + " at path.");
+                }
+            }
```
MarkMichaelis: The logic for this if statement is hard to follow. I would prefer something like:
```
            if (!File.Exists(filename))
            {
                filename = Path.Combine(Path.GetTempPath(), filename)))
            }  
```
This would rely on `File.OpenText(filename)` to throw a `FileNotFoundException` which is all you do anyway. (Notice I don't introduce the `path` variable but that you can if you think it is clearer.)  Also, I am using your logic to check the temp directory although that would be relatively unusual IMO.
`

SecretSanta/SecretSanta.Import/Services/GiftImportService.cs
```diff
@@ -0,0 +1,90 @@
...
+            return wl;
+        }
+
+        public static User GetUserFromHeader(string header)
+        {
+            if (string.IsNullOrEmpty(header))
+                return null;
+            if (!header.StartsWith("Name: "))
+                return null;
+
```
MarkMichaelis: Would a null `User` be a valid return?  Why not throw an exception?

SecretSanta/SecretSanta.Import/Services/GiftImportService.cs
```diff
@@ -0,0 +1,90 @@
...
+                path = filename;
+
+            List<Gift> gifts = new List<Gift>();
+            User user = new User();
+
+            using (StreamReader sr = File.OpenText(path))
+            {
+                string line;
+                int order = 1;
+
```
MarkMichaelis: Why instantiate a User?  You may not have a valid header?

SecretSanta/SecretSanta.Import/Services/GiftImportService.cs
```diff
@@ -0,0 +1,90 @@
...
+{
+    public class GiftImportService
+    {
+        public static Wishlist ReadWishlist(string filename)
+        {
+            if (string.IsNullOrEmpty(filename))
+                throw new ArgumentNullException("Filename cannot be null or empty.");
+
+            string path = Path.Combine(System.IO.Path.GetTempPath() + @"\", filename);
+
```
MarkMichaelis: Is it really valid to return a `WishList` with no `User` and/or no `Gifts`?

SecretSanta/SecretSanta.Import/Services/GiftImportService.cs
```diff
@@ -0,0 +1,90 @@
...
+                while((line = sr.ReadLine()) != null)
+                {
+                    if(!string.IsNullOrEmpty(line) && !string.IsNullOrWhiteSpace(line))
+                    {
+                        gifts.Add(new Gift
+                        {
+                            Title = line,
+                            OrderOfImportance = order
+                        });
```
MarkMichaelis: Can't you just have `if(!string.IsNullOrWhiteSpace(line))`?

SecretSanta/SecretSanta.Import/Services/GiftImportService.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            string[] name;
+
+            string[] headerSplit = header.Trim().Split(':');
+
+            string nameTrimmed = headerSplit[1].Trim();
+
+            if (header.Contains(","))
+                name = nameTrimmed.Split(',');
+            else
```
MarkMichaelis: What if there is no semicolon?

SecretSanta/SecretSanta.Import/Services/GiftImportService.cs
```diff
@@ -0,0 +1,90 @@
+﻿using System;
+using System.IO;
+using System.Collections.Generic;
+using SecretSanta.Domain.Models;
+
+namespace SecretSanta.Import.Services
+{
+    public class GiftImportService
```
MarkMichaelis: If all your methods are static, the class should be static.

SecretSanta/SecretSanta.Import.Tests/ServiceTests/GiftImportServiceTests.cs
```diff
@@ -0,0 +1,196 @@
...
+        public void InitializeGiftImportTests()
+        {
+            InitializeTestFile1();
+            InitializeTestFile2();
+            InitializeTestFile3();
+        }
+
+        private void InitializeTestFile1()
+        {
+            string filepath = System.IO.Path.GetTempPath() + @"\TestFile_NoCommaName.txt";
```
MarkMichaelis: `Path.Combine()` recommended.

SecretSanta/SecretSanta.Import.Tests/ServiceTests/GiftImportServiceTests.cs
```diff
@@ -0,0 +1,196 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SecretSanta.Domain.Models;
+using SecretSanta.Import.Services;
+using System;
+using System.IO;
+
+namespace SecretSanta.Import.Tests.ServiceTests
+{
+    [TestClass]
+    public class GiftImportServiceTests
```
MarkMichaelis: The test structure here is less than ideal IMO.  In every test you initizalize 3 files but don't necessarily use 3 files.  
Also, you have the data you are testing and the data you are inserting retyped. You may as well do the InitializeTestFile calls from each file and refactor the InitializeTestFile methods into one method.
The key issue is you have lots of logic that looks almost the same but the code is not refactored.

PR #39, opened 2019-01-22
Assignment2

grantdwoods:


grantdwoods:
No unit tests for GiftImportService are ready at this time.

SecretSanta/SecretSanta.Import.Tests/Services/GiftImportServiceTest.cs
```diff
@@ -0,0 +1,58 @@
...
+        [TestInitialize]
+        public void SetUpDirectory()
+        {
+            string tempDirPath = Path.GetTempPath();
+            dirPath = Path.Combine(tempDirPath + "SecretSantaGiftImport");
+
+            if (Directory.Exists(dirPath))
+                CleanUpDirectory();
+
+            Directory.CreateDirectory(dirPath);
```
KennethWhite: I like that you created a sub directory for your test data, I should have done this myself.

SecretSanta/SecretSanta.Import.Tests/Services/NameParsingServiceTest.cs
```diff
@@ -0,0 +1,96 @@
...
+            string[] names = giftImportService.ParseHeader(": Woods, Grant");
+        }
+
+        [TestMethod]
+        [DataRow("Name:")]
+        [DataRow("Name:           ")]
+        [DataRow("Name: Grant")]
+        [DataRow("Name: Woods,")]
+        [DataRow("Name:  , Grant")]
+        [ExpectedException(typeof(ArgumentException), "first or last name missing.")]
```
KennethWhite: Nice use of data rows here to test the same exception.

SecretSanta/SecretSanta.Import/Services/NameParsingService.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Import.Services
+{
+    public class NameParsingService
```
KennethWhite: Your NameParsingSevice is really modular, and you were very thorough with addressing edge cases

SecretSanta/SecretSanta.Import.Tests/Services/GiftImportServiceTest.cs
grantdwoods: I thought it was a good idea too! Though I should probably change how I deal with it based on what Mark said in class. Currently if that directory exists I just wipe it out. Mark said "What if someone else had a directory with that name there already". Very unlikely, but I understand his point!

SecretSanta/SecretSanta.Import/Services/NameParsingService.cs
```diff
@@ -0,0 +1,95 @@
...
+        
+            string fullName = ParseFullName(header);
+            string[] names = SeparateFirstAndLast(fullName);
+
+            ValidateNamesArray(names);
+
+            return names;
+        }
+
+        private void ValidateHeaderStart(string header)
```
tdang2018: I really like the way you check whether the header is valid or not by using that method. It is super clearly to understand.

SecretSanta/SecretSanta.Import/Services/NameParsingService.cs
```diff
@@ -0,0 +1,95 @@
...
+            {
+                throw new ArgumentException(nameof(header),
+                    "Header must start with \"Name: \"");
+            }
+        }
+        private void ValidateNamesArray(string[] names)
+        {
+            bool isValid = true;
+
+            if(names.Length !=2 )
```
tdang2018: I think here you can throw Exception if the length of names are not equal to 2 instead of set isValid to false. Then after that, you check if iValid is false, then throw Exception.

SecretSanta/SecretSanta.Import/Services/NameParsingService.cs
```diff
@@ -0,0 +1,95 @@
...
+            return fullName.Trim();
+        }
+        private string[] SeparateFirstAndLast(string fullName)
+        {
+            string[] names;
+
+            if (fullName.Contains(','))
+            {
+                names = fullName.Split(',');
+                SwapToFirstLastFormat(names);
```
tdang2018: you call method SwapToFirstLastFormat() because I think we have 2 format Types:
Name: <FirstName> <LastName>
Name: <LastName>, <FirstName>
Is that correct?

SecretSanta/src/SecretSanta.Domain/Models/ApplicationDbContext.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+
+        }
+
+        protected override void OnModelCreating(ModelBuilder modelBuilder)
+        {
+            modelBuilder.Entity<UserGroup>()
+                .HasKey(userGroup => new { userGroup.UserId, userGroup.GroupId });
+
+            modelBuilder.Entity<UserGroup>()
```
tdang2018: I do not understand the many to many for lamda expression. can you please explain about that? Thanks Grant

SecretSanta/SecretSanta.Import.Tests/Services/GiftImportServiceTest.cs
```diff
@@ -0,0 +1,68 @@
...
+    public class GiftImportServiceTest
+    {
+        string dirPath;
+        string tmpFilePath;
+
+        [TestInitialize]
+        public void SetUpDirectory()
+        {
+            string tempDirPath = Path.GetTempPath();
+            dirPath = Path.Combine(tempDirPath + "SecretSantaGiftImport");
```
tdang2018: What SecretSantaGiftImport in this case mean? Is a file or Directory?

SecretSanta/SecretSanta.Import.Tests/Services/GiftImportServiceTest.cs
```diff
@@ -0,0 +1,68 @@
...
+            ICollection<Gift> gifts = giftImportService.ImportGifts(tmpFilePath);
+            List<Gift> wishList = new List<Gift>(gifts);
+            Assert.AreEqual<string>("Grant", wishList[0].User.FirstName);
+            Assert.AreEqual<string>("Woods", wishList[1].User.LastName);
+            Assert.AreEqual<string>("Gift3", wishList[2].Title);
+
+            File.Delete(tmpFilePath);
+        }
+        [TestMethod]
+        public void ImportGifts_InvalidFile_ArgumentException()
```
tdang2018: This method is empty?

SecretSanta/SecretSanta.Import/Services/NameParsingService.cs
grantdwoods: Exactly! I want to make sure that my array always contains [firstname, lastname] so I found the swap necessary.

SecretSanta/SecretSanta.Import.Tests/Services/GiftImportServiceTest.cs
grantdwoods: I use dirPath to create the directory in the temp folder.

SecretSanta/SecretSanta.Import.Tests/Services/GiftImportServiceTest.cs
grantdwoods: This was just a work in progress, pretty just a place holder at the ti.

SecretSanta/SecretSanta.Import/Services/NameParsingService.cs
grantdwoods: Yeah I was thinking about trying not to repeat the same code too many times, but I agree the bool is probably redundant.

SecretSanta/src/SecretSanta.Domain/Models/ApplicationDbContext.cs
grantdwoods: I do not remember the article I used, but Stokes and a fellow student both posted links on gitter for this. This one might help you out: https://blog.oneunicorn.com/2017/09/25/many-to-many-relationships-in-ef-core-2-0-part-1-the-basics/

SecretSanta/SecretSanta.Import/Services/NameParsingService.cs
```diff
@@ -0,0 +1,95 @@
...
+            }
+            else
+            {
+                names = fullName.Split(" ");
```
MarkMichaelis: Your approach wouldn't work if I had "Name: Grant       Woods" but it would have been easy to allow and would be more robust against silly user errors.

SecretSanta/SecretSanta.Import.Tests/Services/NameParsingServiceTest.cs
```diff
@@ -0,0 +1,101 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SecretSanta.Import.Services;
+using System;
+
+namespace SecretSanta.Import.Tests
+{
+    [TestClass]
+    public class NameParsingServiceTest
+    {
+        public NameParsingService CreateDefaultImporter()
```
MarkMichaelis: I'm not sure you gain much by having a method that only calls a constructor?  It forces the reader to look up what the function does rather than just inline it and see it.  Obviously, if there was or was likely to be more complexity, perhaps.
An alternative approach would be to have a NamingParsingService property and to set it in TestInitialize.

SecretSanta/SecretSanta.Import.Tests/Services/GiftImportServiceTest.cs
```diff
@@ -0,0 +1,126 @@
...
+            DirectoryInfo dirInfo = new DirectoryInfo(dirPath);
+            foreach (FileInfo file in dirInfo.GetFiles())
+            {
+                File.Delete(file.FullName);
+            }
+            Directory.Delete(dirPath);
+        }
+
+        [TestMethod]
+        [DataRow(new[] { "Name: Grant Woods", "Gift1", "Gift2", "Gift3" })]
```
MarkMichaelis: I like the use of the DataRow attribute.

SecretSanta/SecretSanta.Import.Tests/Services/GiftImportServiceTest.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            GiftImportService giftImportService = new GiftImportService();
+
+            File.WriteAllLines(tmpFilePath, fileInputArray);
+            try
+            {
+                ICollection<Gift> gifts = giftImportService.ImportGifts(tmpFilePath);
+                Assert.Fail();
+            }
+            catch (Exception exception)
```
MarkMichaelis: Again... no, just catch an ArgumentException.  
Catching `Exception` and not re-throwing is almost always wrong.  It suggests you could appropriately handle any exception which is unlikely in general.

SecretSanta/SecretSanta.Import.Tests/Services/GiftImportServiceTest.cs
```diff
@@ -0,0 +1,126 @@
...
+            Assert.AreEqual<string>("Woods", wishList[1].User.LastName);
+            Assert.AreEqual<string>("Gift3", wishList[2].Title);
+
+            File.Delete(tmpFilePath);
+        }
+
+        [TestMethod]
+        [DataRow(new[] { "Name: Grant Woods", "", "", "" })]
+        [DataRow(new[] { "Name: Grant Woods", "            ", " ", "" })]
+        public void ImportGifts_ValidHeaderMissingGifts_ThrowsArgumentException(string[] fileInputArray)
```
MarkMichaelis: The code os convoluted.  Using an ExpectedException attribute would help significantly.

MarkMichaelis:
Grant.... I really like your implementation.  Nice job!!

PR #41, opened 2019-01-23
Assignment2

Isaac-Bliss:
Assignment2


Isaac-Bliss:
Had a lot of trouble getting to the point where I could actually code instead of fighting against GitHub and .yml, so there's really nothing here at this point. Just testing basic file stuff thus far.

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Models
+{
+    public class Gift : Entity
```
MarkMichaelis: Wait, no validation?  Surely you should have validation I would think.

MarkMichaelis:
Overall, nice job.  The biggest things were the fact that you never actually imported to the database, never leveraged your Domain classes (such as User and Gift). 
Another important consderation would be to not require your API to be called in a specific order.  This isn't enforcable through the API interface leading to errors by programmers using your API.

PR #42, opened 2019-01-23
Assignment2 - Brad Howard

BHoward85:
Unit Test for Import Services, Code coverage is close to 93% overall, and for 98% import services unit test.


SecretSanta/test/SecretSanta.Domain.Tests/ImportServiceTest.cs
```diff
@@ -0,0 +1,156 @@
...
+            importService.Dispose();
+        }
+
+        [TestMethod]
+        public void ParseName_FirstNameFirst()
+        {
+            string path = Path.Combine(System.IO.Path.GetTempPath(), "inputFile.txt");
+            ImportService importService = new ImportService(path);
+
+            Assert.AreEqual<string>("Brad", importService.ParseName(importService.ReadName())[0]);
```
RichardTeller: For the ParseName tests, in theory all we should be doing is seeing if the ParseName method works, that is, if it can take an input string and parse out the name.  To make the test more of a unit test, you should be able to take out ParseName(importService.ReadName()), and just pass it a string instead.  Because otherwise at that point you're testing both ParseName and ReadName.  And yes sometimes you can't get away with testing just the one method, but consider changing it around so its just calling ParseName if you have the time.

SecretSanta/src/SecretSanta.Import/ImportService.cs
```diff
@@ -0,0 +1,106 @@
...
+        public ImportService(string path)
+        {
+            if (File.Exists(path))
+            {
+                Istream = new FileStream(path, FileMode.Open, FileAccess.Read);
+                InstanceCount++;
+            }
+            else
+            {
+                throw new FileNotFoundException("File does not exist error");
```
joeriddles: Nice use of the specific type of error `FileNotFoundException`.

SecretSanta/src/SecretSanta.Import/ImportService.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        ~ImportService()
+        {
+            Dispose();
+        }
+
+        public string ReadName()
+        {
+            StreamReader streamReader = new StreamReader(Istream);
+            string name = streamReader.ReadLine();
```
joeriddles: Declaring `name` seems unnecessary. You could just return `streamReader.ReadLine();`.

SecretSanta/src/SecretSanta.Import/ImportService.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            return name;
+        }
+
+        public string[] ParseName(string word)
+        {
+            bool hasComma = false;
+            string[] name = new string[2];
+            string[] temp;
```
joeriddles: This could be declared on the same line as `name`. Not sure if that violates a coding standard without checking, but food for thought.

SecretSanta/src/SecretSanta.Import/ImportService.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            StreamReader streamReader = new StreamReader(Istream);
+            string name = streamReader.ReadLine();
+
+            return name;
+        }
+
+        public string[] ParseName(string word)
+        {
+            bool hasComma = false;
```
joeriddles: I believe `bool` defaults to false.

SecretSanta/src/SecretSanta.Import/ImportService.cs
```diff
@@ -0,0 +1,106 @@
...
+            string[] name = new string[2];
+            string[] temp;
+
+            if (word.Contains(":"))
+            {
+                temp = word.Split(':');
```
joeriddles: I'm a fan of removing the brackets from single line clauses for `if` statements. I think this is just more of a stylistic choice.

SecretSanta/src/SecretSanta.Import/ImportService.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                throw new FormatException("Missing first or last name.");
+            }
+
+            temp[0] = temp[0].Trim();
+            temp[1] = temp[1].Trim();
+
+            if (hasComma)
+            {
+                name[0] = temp[1];
```
joeriddles: This `if/else` seems unnecessary. This code could be moved lines 59 and 63, no?

SecretSanta/test/SecretSanta.Domain.Tests/ImportServiceTest.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            BuildTestFile("Name: Brad Howard", "inputFile.txt");
+            BuildTestFile("Name: Howard, Brad", "inputFile2.txt");
+            BuildTestFile("Name:", "inputFile3.txt");
+            BuildTestFile("Name", "inputFile4.txt");
+            BuildTestFile("Name: Howard", "inputFile5.txt");
+            BuildTestFile("Name: Howard,", "inputFile6.txt");
+            BuildTestFile("Name:Howard,", "inputFile7.txt");
+        }
+
```
joeriddles: I don't see a `[TestCleanup]`. Do these files ever get deleted?

SecretSanta/test/SecretSanta.Domain.Tests/ImportServiceTest.cs
```diff
@@ -0,0 +1,156 @@
...
+            }
+        }
+
+        [TestMethod]
+        public void OpenFileForImport_CanRead()
+        {
+            string path = Path.Combine(System.IO.Path.GetTempPath(), "inputFile.txt");
+            ImportService importService = new ImportService(path);
+            
+            Assert.AreEqual<bool>(true, importService.Istream.CanRead);
```
joeriddles: Could replace `Assert.AreEqual<bool>(true, ...)` with `Assert.IsTrue(...)`

joeriddles: Good job on remembering `<bool>` 👍  
I think I forgot that in some places in my tests.

SecretSanta/src/SecretSanta.Import/ImportService.cs
```diff
@@ -0,0 +1,106 @@
...
+                name[1] = temp[0];
+            }
+            else
+            {
+                name = temp;
+            }
+
+            return name;
+        }
```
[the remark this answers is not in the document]
Cwhite534: I think returning a tuple representing the name, or a user object might be better than returning an array. I also agree that there is more logic than needed. for example, the last if/else block could be simplified to `return hascomma ? Array.Reverse(temp) : temp;` This assumes that temp has been checked and only contains 2 items, the first name, and the last.

SecretSanta/test/SecretSanta.Domain.Tests/ImportServiceTest.cs
BHoward85: There was a thread lock on the files, I tried to call the delete on them and the system threw a exception about them being used by another thread.

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Models
+{
+    public class Gift : Entity
```
MarkMichaelis: Why no validation on your properties?

SecretSanta/SecretSanta.sln
```diff
@@ -0,0 +1,74 @@
...
+MinimumVisualStudioVersion = 15.0.26124.0
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{9E2B84C6-9650-4109-BC73-502EB4FB1132}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SecretSanta.Domain", "src\SecretSanta.Domain\SecretSanta.Domain.csproj", "{A605EAB5-FC10-4255-B617-38C011447FAA}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{0165C384-B8D1-4750-81A5-A71B2942B491}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SecretSanta.Domain.Tests", "test\SecretSanta.Domain.Tests\SecretSanta.Domain.Tests.csproj", "{070E74C2-C6D3-421F-8628-8AFFE6DA837F}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SecretSanta.Import", "src\SecretSanta.Import\SecretSanta.Import.csproj", "{CB991578-3182-40E6-927B-93277072127A}"
```
MarkMichaelis: Since you have an Import assembly, I would like to have seen in Import.Tests project.

SecretSanta/src/SecretSanta.Import/ImportService.cs
```diff
@@ -0,0 +1,106 @@
+﻿using System;
+using System.IO;
+
+namespace SecretSanta.Import
+{
+    public class ImportService : IDisposable
```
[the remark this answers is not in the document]
MarkMichaelis: Agreed... and you remembered to implement the finalizer along with it (which several other people in class did not.)

SecretSanta/src/SecretSanta.Import/ImportService.cs
MarkMichaelis: Sure, but he may be checking it with the theoretical possibility it hasn't been set, thus issuing a warning so setting it avoids the warning.

SecretSanta/src/SecretSanta.Import/ImportService.cs
MarkMichaelis: I prefer it this way.  Especially since they one is initialized and the other isn't.
I rarely declare mutliple variables on the same line.

MarkMichaelis:
Overall I think you did a nice job.
The one thing to notice about your API is that a caller could invoke your methods in any order and you can't control it in the API.  Perhaps this woudl be easier with just one public method exposed.

PR #43, opened 2019-01-23
Assignment2

RichardTeller:


SecretSanta/test/SecretSanta.Import.Tests/UserGiftImporterTests.cs
```diff
@@ -13,15 +15,15 @@ public class UserGiftImporterTests
         [TestInitialize]
         public void TestInitialize()
         {
-            TestFilePath = Environment.CurrentDirectory + "\\TestFile.txt";
+            TestFilePath = System.IO.Path.GetTempFileName();
```
BHoward85: I had some trouble with the file placement and access. Now seeing how you used the Temp File and ease of clean up. Very good.

SecretSanta/src/SecretSanta.Import/UserGiftImporter.cs
```diff
@@ -0,0 +1,127 @@
...
+
+                    return new string[] { headerType, firstName, lastName };
+                }
+                else
+                {
+                    string firstName = header.Substring(0, header.IndexOf(" "));
+                    string lastName = header.Substring(header.IndexOf(" ") + 1);
+
+                    return new string[] { headerType, firstName, lastName };
+                }
```
BHoward85: Have you tested all the cases that should there be a missing first name, last name, or that there is no name past the ":" in the header? I just letting you know that so far I can not see that you are checking for missing parts.

RichardTeller: I don't believe we are supposed to check for extraneous things like that.  Mark said the point of the assignment was to practice TDD.  In the assignment instructions it says we are given a valid file.  The only things we need to check for in terms of format is which of the 2 header formats is the header, and the gifts can contain blank lines or if its not blank then its a valid gift name.

MarkMichaelis:
Overall I think you did a nice job!
The API design could be improved.  You never actually imported the data you parsed which was seemingly strange?

PR #45, opened 2019-01-23
Assignment2

tdang2018:
Pull Request for Assignment 2[First Draft]


SecretSanta/SecretSanta.Import/WishListFile.cs
```diff
@@ -0,0 +1,91 @@
...
+                    //Read Text File into String Array 
+                    lines = File.ReadAllLines(fileName);
+                }                
+            }
+            return lines;
+        }
+     
+        public  bool IsValidHeader(string header)
+        {
+            if (string.IsNullOrEmpty(header))
```
efueng: You might consider checking for than null when you validate a header. Headers follow a specific format (you can find it in Assignment2.md)

SecretSanta/SecretSanta.Import/WishListFile.cs
```diff
@@ -0,0 +1,91 @@
...
+        public string[] OpenFile(string fileName)
+        {
+            string[] lines;
+            if (fileName == null)
+            {
+                throw new ArgumentNullException("Filename is null");
+            }
+            else
+            {
+               // string path = System.Environment.CurrentDirectory;
```
efueng: Usual warnings about commented out code

SecretSanta/SecretSanta.Import/WishListFile.cs
```diff
@@ -0,0 +1,91 @@
+﻿using System;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace SecretSanta.Import
+{
+    public class WishListFile
+    {
+        public string CreateTempFile()
+        {
```
efueng: Not sure you need to create temporary files inside this class.

SecretSanta/SecretSanta.Import/WishListFile.cs
```diff
@@ -0,0 +1,91 @@
...
+     
+        public  bool IsValidHeader(string header)
+        {
+            if (string.IsNullOrEmpty(header))
+            {
+                return false;
+            }
+            else
+            {
+                string[] temp = header.Split(":");
```
efueng: This code seems to be here strictly to assist in testing, which seems like a bad idea. I think you want to separate the two concerns.

SecretSanta/SecretSanta.Import/WishListFile.cs
```diff
@@ -0,0 +1,91 @@
...
+                if (header.ToLower().Contains(", "))
```
efueng: This is a little pedantic, but you don't need the call to ToLower() since you're just checking for a comma.

SecretSanta/SecretSanta.Import/WishListFile.cs
efueng: Consider using a more descriptive name than temp. It helps with readability and intellisense makes using long descriptive names relatively easy.

SecretSanta/SecretSanta.Import/WishListFile.cs
```diff
@@ -0,0 +1,91 @@
...
+                {
+                    
+                    return temp[1] == " Tuan, Dang" ? true : false;
+                }
+                else
+                {
+                    return temp[1] == " Tuan Dang" ? true : false;
```
efueng: I think right now IsValidHeader will only return true if it's your name in the header. You're going to want to expand this so it can validate headers with different names.

SecretSanta/SecretSanta.Import.Tests/WishListFileTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            WishListFile = new WishListFile();
+            //use to create test file
+            TempFileName = WishListFile.CreateTempFile();
+            //Lines = WishListFile.OpenFile("wishlistuser.txt");           
+            Lines = WishListFile.OpenFile(TempFileName);
+        }
+       
+        [TestMethod]
+        public void CheckHeader_FirstNameSpaceLastName_ReturnsTrue()
```
efueng: Your tests and your IsValidHeader method are very dependent on each other. For example, your tests only seem to check against variations of your name. Normally this would be ok, but your name is also hardcoded into some checks in your IsValidHeader method. The same is true for many of your other tests.

SecretSanta/SecretSanta.Import.Tests/WishListFileTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        public WishListFile WishListFile=null;
+        public string TempFileName;
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            WishListFile = new WishListFile();
+            //use to create test file
+            TempFileName = WishListFile.CreateTempFile();
+            //Lines = WishListFile.OpenFile("wishlistuser.txt");
```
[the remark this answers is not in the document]
tdang2018: @efueng : Yeah Thanks Eric.

SecretSanta/SecretSanta.Import.Tests/WishListFileTests.cs
tdang2018: @efueng : You are correct. Because I just follow the TDD rules: Write the simplest code to test first then refactor. I am working on it.

SecretSanta/SecretSanta.Import/WishListFile.cs
tdang2018: Me neither. Because I just wanna any method related to File will be put into that class. But I will change later.

SecretSanta/SecretSanta.Import/WishListFile.cs
tdang2018: @efueng : I have question for you . So we cannot comment out on code, How about the comment out the pseucode? Can I?

SecretSanta/SecretSanta.Import/WishListFile.cs
tdang2018: @efueng : Thanks Eric.

SecretSanta/SecretSanta.Import/WishListFile.cs
tdang2018: @efueng : Yeah. I will do it now. Thanks

SecretSanta/SecretSanta.Import/WishListFile.cs
tdang2018: @efueng because C# is case sensitive, So I just want to ignore that by comparing Firstname or lastname. But you are right, I will change it. .Thanks

SecretSanta/SecretSanta.Import.Tests/WishListFileTests.cs
```diff
@@ -0,0 +1,93 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SecretSanta.Import.Tests
+{
+    [TestClass]
+    public class WishListFileTests
+    {
+        public string[] Lines;
+        public WishListFile WishListFile=null;
```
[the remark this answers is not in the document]
tdang2018: Thanks Eric. I will do it now

SecretSanta/SecretSanta.Import/WishListFile.cs
tdang2018: Yeah. I will change now.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text.RegularExpressions;
+
+namespace SecretSanta.Import
+{
+    public class WishListFileParser : IDisposable
```
MarkMichaelis: Great job noticing and implementing IDisposable!  However, you neglected to implement the finalizer that accompanies `Dispose()` for when people forget to call dispose (as I belive you did in your tests:)

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        public StreamReader StreamReader { get; set; }
```
MarkMichaelis: I suspect this shouldn't have a public setter?  You don't want your caller changing the stream reader I don't think.  If nothing else, you wouldn't know what state it was in.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+            return giftNames;
+        }    
+            
+        public  bool IsFileExists(string curFile)
+        {
+            return File.Exists(curFile) ? true : false;
+        }
+
+        public void Dispose()
+        {
```
MarkMichaelis: I would think the implementation here would just invoke `CloseImportFile`?

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+            if (StreamReader != null)
+            {
+                StreamReader.Dispose();
```
MarkMichaelis: It would seem that `StreamReader.?Dispose()` would be clear and simpler here.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+                StreamReader.Close();
+                StreamReader = null;
+            }
+        }
+     
+        public string[] SplitHeader(string header)
+        {
+            if(header == null)
+            {
+                throw new ArgumentNullException();
```
MarkMichaelis: You should pass specifiy `nameof(header)` (or whatever the null argument is) in all cases when throwing an Argument(Null)Exception.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+            else if(header ==String.Empty)
+            {
+                throw new ArgumentException("Header is empty");
```
MarkMichaelis: Again... specifiy the argument name using `nameof`.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+                    StreamReader = new StreamReader(fileName);
+                }
+            }
+        }
+
+        public void CloseImportFile()
+        {
+            if(StreamReader != null)
+            {
+                StreamReader.Close();
```
MarkMichaelis: How do you determine when to call `CloseImportFile()` and `Dispose()`?  Consier combining these methods.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                throw new ArgumentNullException();
+            }
+            else if(header ==String.Empty)
+            {
+                throw new ArgumentException("Header is empty");
+            }
+            else
+            {
+                header = header.Trim();
```
MarkMichaelis: Glad you remembered `Trim()`

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+                if(!header.Contains(":") || !header.StartsWith("Name"))
```
MarkMichaelis: You check for ':' multiple times.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+                {
+                    throw new ArgumentException("Invalid Header!");         
+                }
+                string headerNameWithColon = header.Substring(0, header.IndexOf(":"));
```
MarkMichaelis: For example, even though you checked with `Contains()` you check twice here using `IndexOf()`.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+                string remainedHeader = header.Substring(header.IndexOf(":") + 1);
+                remainedHeader = remainedHeader.Trim();
+                //Check format <LastName>, <FirstName>
+                if(remainedHeader.Contains(","))
```
MarkMichaelis: Similarly... you check for comma muliple times.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+        public void CloseImportFile()
+        {
+            if(StreamReader != null)
+            {
+                StreamReader.Close();
+                StreamReader = null;
+            }
+        }
+     
+        public string[] SplitHeader(string header)
```
MarkMichaelis: This seems a suboptimal return type?  Can you return the User or, at a minimum, a tuple?  Any time you return an array where each element represents a different category of data it becomes non-obvious which element in the arrage is for what.

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Models
+{
+    //hold the title, order of importance, url, description, and User
+    public class Gift : Entity
```
MarkMichaelis: Ughhhh!!! Why no property validation?

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+                    if(names.Length != 2)
+                    {
+                        throw new ArgumentException("Missing first name or last name!");
+                    }
+                    string lastName = remainedHeader.Substring(0, remainedHeader.IndexOf(","));
+                    string firstName = remainedHeader.Substring(remainedHeader.IndexOf(",") + 2);              
+
+                    return new string[] { headerNameWithColon, firstName, lastName };
+                }
+                else //<FirstName> <LastName>
```
MarkMichaelis: It would seem that som refactoring could be done.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+                    }
+
+                    string firstName = remainedHeader.Substring(0, remainedHeader.IndexOf(" "));
+                    string lastName = remainedHeader.Substring(remainedHeader.IndexOf(" ") + 1);                  
+                    return new string[] { headerNameWithColon, firstName, lastName };
+                }              
+            }
+        }
+
+        public List<string> ReadListOfGifts()
```
MarkMichaelis: Perhaps my biggest concern with your code is that callers would have to know what order to call methods in.  Do you really want them to be ale to call in different order - or, perhaps, skip the invocation of some?  I doubt it.

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+                        throw new ArgumentException("Missing first name or last name!");
+                    }
+
+                    string firstName = remainedHeader.Substring(0, remainedHeader.IndexOf(" "));
+                    string lastName = remainedHeader.Substring(remainedHeader.IndexOf(" ") + 1);                  
+                    return new string[] { headerNameWithColon, firstName, lastName };
+                }              
+            }
+        }
+
```
MarkMichaelis: Why return `List<Gift>` rather than `IEnumerable<Gift>`?  Would you want the caller indexing into the list or changing the list?

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+{
+    public class WishListFileParser : IDisposable
+    {
+        public StreamReader StreamReader { get; set; }            
+
+        public void OpenImportFile(string fileName)
+        {
+            if(fileName == null)
+            {
+                throw new ArgumentNullException("File is null");
```
MarkMichaelis: Ughh.... here you specify the parameter name is "File is null"?

SecretSanta/SecretSanta.Import/WishListFileParser.cs
```diff
@@ -0,0 +1,117 @@
...
+                    throw new ArgumentException("Invalid Header!");         
+                }
+                string headerNameWithColon = header.Substring(0, header.IndexOf(":"));
+                string remainedHeader = header.Substring(header.IndexOf(":") + 1);
+                remainedHeader = remainedHeader.Trim();
+                //Check format <LastName>, <FirstName>
+                if(remainedHeader.Contains(","))
+                {
+                    string[] names = remainedHeader.Split(",");
+                    if(names.Length != 2)
```
MarkMichaelis: What if there were multiple spaces?  You wouldn't allow that?  You use `Trim()` earlier but here you wouldn't support.

SecretSanta/SecretSanta.Import.Tests/WishListFileParserTests.cs
```diff
@@ -0,0 +1,199 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace SecretSanta.Import.Tests
+{
+    [TestClass]
+    public class WishListFileParserTests
+    {        
+        public WishListFileParser WishListFileParser { get; set; }
```
MarkMichaelis: Not a major deal but notice you support `IDisposable` for WishListFileParser but you never call `Dispose()`.

SecretSanta/SecretSanta.Import.Tests/WishListFileParserTests.cs
```diff
@@ -0,0 +1,199 @@
...
+        public string TempFileName { get; set; }
+
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            WishListFileParser = new WishListFileParser();            
+            TempFileName = CreateTempFile();                       
+        }
+
+        public string CreateTempFile()
```
MarkMichaelis: Nice job refactoring this method out.
You could make it null since it doesn't access any instance data.

SecretSanta/SecretSanta.Import.Tests/WishListFileParserTests.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            string fileName = "";
+            try
+            {
+                fileName = Path.GetTempFileName();
+                FileInfo fileInfo = new FileInfo(fileName);
+                fileInfo.Attributes = FileAttributes.Temporary;
```
MarkMichaelis: Good to see the use of this attribute here.

SecretSanta/SecretSanta.Import.Tests/WishListFileParserTests.cs
```diff
@@ -0,0 +1,199 @@
...
+            }
+            catch (Exception ex)
+            {
+                throw new Exception(ex.Message);
+            }
+            return fileName;
+        }
+
+        public void InsertLinesToTempFile(string tempFile, string[] lineToInsert)
```
MarkMichaelis: I liked the way you refactored this out rather than having lots of different files.

SecretSanta/SecretSanta.Import.Tests/WishListFileParserTests.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            try
+            {
+                // Write to the temp file.                
+                StreamWriter streamWriter = File.AppendText(tempFile);
```
MarkMichaelis: StreamWriter is disposable - I suggest leveraging the using statment here.

SecretSanta/SecretSanta.Import.Tests/WishListFileParserTests.cs
```diff
@@ -0,0 +1,199 @@
...
+            using (WishListFileParser = new WishListFileParser())
+            {
+                string[] headers = WishListFileParser.SplitHeader(" Name: Tuan Dang ");
+                Assert.AreEqual<string>("Name", headers[0]);
+                Assert.AreEqual<string>("Tuan", headers[1]);
+                Assert.AreEqual<string>("Dang", headers[2]);
+            }
+        }
+
+        [TestMethod]
```
MarkMichaelis: You could refactor all these ArgumentException tests into Data driven tests, simplifying significant code here.

SecretSanta/SecretSanta.Import.Tests/WishListFileParserTests.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            using (WishListFileParser = new WishListFileParser())
+            {                
+                InsertLinesToTempFile(TempFileName, lines);
+                WishListFileParser.OpenImportFile(TempFileName);
+                List<string> gifts = WishListFileParser.ReadListOfGifts();                
+                Assert.AreEqual<string>("Gift1", gifts[0]);
+                Assert.AreEqual<string>("Gift2", gifts[1]);
+                Assert.AreEqual<string>("Gift3", gifts[2]);
+            }
```
MarkMichaelis: What if an exception is thrown - wouldn't DeleteTempFile() never get called?  Fortuantley, you call `DeleteTempFile()` from TestCleanup - begging the question of why you explicitly delete it here at all?

SecretSanta/SecretSanta.Import.Tests/WishListFileParserTests.cs
```diff
@@ -0,0 +1,199 @@
...
+            WishListFileParser = new WishListFileParser();            
+            TempFileName = CreateTempFile();                       
+        }
+
+        public string CreateTempFile()
+        {
+            string fileName = "";
+            try
+            {
+                fileName = Path.GetTempFileName();
```
MarkMichaelis: While this works, if you are debugging and your TestCleanup never runs, you won't have a way to delete the file.  Consider adding some extension or such so that you can delete the left-around-files on startup.

MarkMichaelis:
Overall you did a nice job.  On additional point to mention is that you never actually imported the data into the database. You only parsed it.

PR #46, opened 2019-01-23
Assignment2

joeriddles:
# Relevant new files:
## [ImportTests.cs](https://github.com/IntelliTect-Samples/EWU-CSCD496-2019-Winter/compare/Assignment2...joeriddles:Assignment2?expand=1#diff-f1f88a8ba5a1a35537e30614b0f2d2fd)


SecretSanta/src/SecretSanta.Import/Import.cs
```diff
@@ -0,0 +1,58 @@
...
+	public class Import
+	{
+		public string Filename { get; }
+
+		public Import(string filename)
+		{
+			if (filename is null)
+				throw new ArgumentNullException(nameof(filename));
+			if (!File.Exists(filename))
+				throw new ArgumentException(nameof(filename), "File does not exist.");
```
COsborn2: I think the ArgumentException is message, paramName. So I think you need to flip your parameters. More on that [here](https://docs.microsoft.com/en-us/dotnet/api/system.argumentexception.-ctor?view=netframework-4.7.2#System_ArgumentException__ctor_System_String_System_String_)

SecretSanta/src/SecretSanta.Import/Import.cs
```diff
@@ -0,0 +1,58 @@
...
+		{
+			string header = File.ReadLines(Filename).First();
+			string[] headerArray = header.Split(':');
+			if (headerArray.Length != 2 || !headerArray[0].Equals("Name"))
+				throw new ArgumentException(nameof(Filename), "Header formatted incorrectly.");
+
+			string name = headerArray[1].Trim();
+			string[] nameFormat = name.Split(',', 2);
+
+			if (nameFormat.Length == 1)
```
COsborn2: These two back to back if statements is a good opportunity to transition to using a switch statement. Since its just two ifs its not really a big deal, but its just something you can consider. It would make code maintainability easier in the future, if (theoretically) you can think of a scenario where you'll have additional items in your nameFormat array.

SecretSanta/src/SecretSanta.Import/Import.cs
```diff
@@ -0,0 +1,58 @@
...
+		}
+
+		public List<Gift> ReadBody(User user)
+		{
+			if (user is null)
+				throw new ArgumentNullException(nameof(user), "Cannot add gifts before user has been created.");
+
+			List<Gift> gifts = new List<Gift>();
+			IEnumerable<string> body = File.ReadLines(Filename).Skip(1);
+			foreach (var line in body.Where(line => line.Trim() != ""))
```
COsborn2: I like the use of Linq here. If you really want to get crazy with it you can drop the foreach and make this all one big Linq statement. I think it a bit harder to read, but its cool. Totally up to you.
You can drop the gifts List declaration and instead of the foreach you can do this
`return body.Where(line => line.Trim() != "").Select(line => new Gift(line.Trim(), "", user.Id)).ToList();`
That should work..I think.

SecretSanta/src/SecretSanta.Import/Import.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using SecretSanta.Domain.Models;
+
+namespace SecretSanta.Import
+{
+	public class Import
+	{
+		public string Filename { get; }
```
COsborn2: I like the non-static Implementation you made here. I can see how having an instance of this type could be more advantageous than a static one.

SecretSanta/test/SecretSanta.Import.Tests/ImportTests.cs
```diff
@@ -0,0 +1,175 @@
...
+		[TestMethod]
+		[ExpectedException(typeof(ArgumentException))]
+		public void Import_fileDoesNotExist_ThrowsException()
+		{
+			Import import = new Import("__thisfiledontexist.fake");
+		}
+
+		[TestMethod]
+		[ExpectedException(typeof(ArgumentException))]
+		public void Import_fileIsEmpty_ThrowsException()
```
COsborn2: You could specify the type of the exception instead of saying ThrowsException. Its not a coding standard or anything, but it would provide more information when viewing tests in the test runner. Certainly not a big deal.

SecretSanta/src/SecretSanta.Import/Import.cs
```diff
@@ -0,0 +1,58 @@
...
+				throw new ArgumentNullException(nameof(user), "Cannot add gifts before user has been created.");
+
+			List<Gift> gifts = new List<Gift>();
+			IEnumerable<string> body = File.ReadLines(Filename).Skip(1);
+			foreach (var line in body.Where(line => line.Trim() != ""))
+				gifts.Add(new Gift(line.Trim(), "", user.Id));
+
+			return gifts;
+		}
+	}
```
COsborn2: You can add a few tests in here for something like extra spaces in header (before Name:, between firstName and lastName, etc.). It depends on how crazy you want to get with it, but showing that your code does, or does not, accept strange spacing scenarios like "Name: John          Smith" would be helpful for others looking at the unit tests.

SecretSanta/test/SecretSanta.Import.Tests/ImportTests.cs
```diff
@@ -0,0 +1,175 @@
...
+				File.Delete("__empty.txt");
+		}
+
+		[TestMethod]
+		public void InitializeAndCleanupAreWorking()
+		{
+			Assert.IsFalse(File.Exists("__empty.txt"));
+			Assert.IsTrue(File.Exists("__temp.txt"));
+			Assert.IsTrue(File.Exists("__temp2.txt"));
+		}
```
COsborn2: Nice way of making absolutely sure that the Initialization is working as expected. If this test exists to make sure you did Initialization correctly, it's great. But if you're using this test as a way of making sure that the files exist before your test executes dont forget that the order the tests run is not guaranteed.

SecretSanta/src/SecretSanta.Import/Import.cs
Keboo: Even better would be to only do the `.Trim()` call once. Something like
```C#
List<Gift> gifts = File.ReadLines(Filename)
    .Skip(1)
    .Select(line => line.Trim())
    .Where(line => != "")
    .Select(line => new Gift(line, "", user.Id))
    .ToList();
```

SecretSanta/src/SecretSanta.Import/Import.cs
joeriddles: This is a good point considering there could potentially be a middle name, etc.

SecretSanta/src/SecretSanta.Import/Import.cs
joeriddles: @COsborn2 and @Keboo, these are both great suggestions that leverage the heck out of LINQ. I'm not always a huge fan of huge, multi-line LINQ statements as I think it reduces the readability to those who are not super familiar with LINQ syntax (I know it was a struggle for me when first learning C#). Either way, I'm implementing Keboo's code as I think it is more elegant than my own solution.

SecretSanta/test/SecretSanta.Import.Tests/ImportTests.cs
joeriddles: Great suggestion 👍

SecretSanta/test/SecretSanta.Import.Tests/ImportTests.cs
joeriddles: Thanks! 
`__temp.txt` and `__temp2.txt` are created by my `ClassInitializer` method and so they should exist regardless of what test is run first. `__empty` is cleaned up before (and attempts to get cleaned up after) every test that is run, so the order the tests are run should not affect this test.

SecretSanta/src/SecretSanta.Import/Import.cs
joeriddles: Added some new tests! 😄

PR #48, opened 2019-01-23
Assignment2

KennethWhite:
Assignment2 initial solution with @Cwhite534


SecretSanta/src/SecretSanta.Library/FileUtility.cs
```diff
@@ -0,0 +1,18 @@
...
+    public class FileUtility
+    {
+        //Could possibly try to detect whether the path is a full path or a partial
+        public static FileStream OpenFile(string fileName)
+        {
+            if (string.IsNullOrWhiteSpace(fileName))
+            {
+                throw new ArgumentException($"Parameter fileName empty or null on call to OpenFile.", fileName);
+            }
+            return new FileStream(fileName, FileMode.Open, FileAccess.Read);
```
raymondshiner: Does this line of code account for the filename being either an absolute path or a relative path?

Cwhite534: It does account for both absolute, and relative paths, but keep in mind the relative path is from wherever the dll is when being ran. In a test for example, `SecretSanta\test\SecretSanta.Library.Tests\bin\Debug\netcoreapp2.2\` would probably be where the dll is. Passing `./test.txt`  as a file would work, if there is a "test.txt" file in the `\netcoreapp2\` folder.

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace SecretSanta.Domain.Models
+{
+    //hold the title, order of importance, url, description, and User
+    public class Gift : Entity
```
MarkMichaelis: No validation?  :(

SecretSanta/src/SecretSanta.Library/FileParser.cs
```diff
@@ -0,0 +1,62 @@
...
+
+
+namespace SecretSanta.Library
+{
+    public class FileParser
+    {
+        public static User ParseWishlistFile(string wishlistFileName)
+        {
+            User user;
+            using (var streamReader = new StreamReader(FileUtility.OpenFile(wishlistFileName)))
```
MarkMichaelis: Glad to see you only read the file once and then process the lines for each type of data sequentially.

SecretSanta/src/SecretSanta.Library/FileParser.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                user = ParseWishlistHeader(streamReader);
+                user.Wishlist = ParseWishlistGifts(streamReader);
+            }
+
+            return user;
+        }
+
+        private static User ParseWishlistHeader(StreamReader fin)
```
MarkMichaelis: It isn't wrong but I would probably have a `string line` argument rather than the `StreamReader`.  It would ensure that the stream couldn't mistakenly be modified (closed, cursor relocated inappropriately, etc.).  This would have been easier with the `System.IO.File.ReadLine()` method I expect.

SecretSanta/src/SecretSanta.Library/FileParser.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            string line;
+            while (string.IsNullOrWhiteSpace(line = fin.ReadLine()))
+            {
+                if (fin.EndOfStream) throw new InvalidDataException("File does not contain a header");
+            }
+
+            if (!line.Contains("Name"))
```
MarkMichaelis: `Contains()` doesn't seem sufficient.  Wouldn't `string.StartsWith()` be preferable. In fact, check for `"Name: "` even?

SecretSanta/src/SecretSanta.Library/FileParser.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            return user;
+        }
+
+        private static User ParseWishlistHeader(StreamReader fin)
+        {
+            string line;
+            while (string.IsNullOrWhiteSpace(line = fin.ReadLine()))
+            {
+                if (fin.EndOfStream) throw new InvalidDataException("File does not contain a header");
```
MarkMichaelis: I really like your choice of exception here.

SecretSanta/src/SecretSanta.Library/FileParser.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            if (!line.Contains("Name"))
+                throw new InvalidDataException($"Header: \"{line}\" is an improper format");
+
+            var swapNames = line.Contains(',');
```
MarkMichaelis: What if the comma is the first character on the line?

SecretSanta/src/SecretSanta.Library/FileParser.cs
```diff
@@ -0,0 +1,62 @@
...
+            var names = line.Split().Skip(1).ToArray();
```
MarkMichaelis: I would suggest trying to be a little more robust.  What if there were two spaces, or no space after "Name:".

SecretSanta/src/SecretSanta.Library/FileParser.cs
```diff
@@ -0,0 +1,62 @@
...
+            if (names.Length != 2)
+                throw new InvalidDataException($"Header: \"{line}\" does not contain a proper user name");
+
+            var firstname = swapNames ? names[1] : names[0];
+            var lastname = swapNames ? names[0].Remove(names[0].Length - 1) : names[1];
+
+            return new User {FirstName = firstname, LastName = lastname};
```
MarkMichaelis: You have no validation on User.  What if the first line was "Name:   " (three spaces)?

SecretSanta/src/SecretSanta.Library/FileUtility.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.IO;
+
+namespace SecretSanta.Library
+{
+    public class FileUtility
+    {
+        public static FileStream OpenFile(string fileName)
```
MarkMichaelis: Hmmm.... I'm not sure what this file is doing that File or StreamReader doesn't do out of the box?  Not sure there is enough value here to warrent writing this.

SecretSanta/test/SecretSanta.Library.Tests/FileParserTests.cs
```diff
@@ -0,0 +1,140 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace SecretSanta.Library.Tests
+{
+    [TestClass]
+    public class FileParserTests
```
MarkMichaelis: High level comments:
* I would like to see more invalid data tests (like ,Name: Inigo Montoya").
* You are reusing the same data but typing it multiple times.  Refactor or use different data so that you have more test scenarios.

MarkMichaelis:
> The code is clean an succinct. Nice job!
> The biggest issue is probably your lack of validation logic and the lack of invalid data tests causes you to miss the lack of validation.

MarkMichaelis:
> > The code is clean an succinct. Nice job!
> > The biggest issue is probably your lack of validation logic and the lack of invalid data tests causes you to miss the lack of validation.

I was also pleased to see you only read the file once.  Using `File.ReadLine` might have been a cleaner/simler implementation.

SecretSanta/src/SecretSanta.Library/FileParser.cs
raymondshiner: Agreed, Didn't think about that enough

SecretSanta/src/SecretSanta.Library/FileParser.cs
raymondshiner: So for this that situation won't occur. Because since i have the condition checking for names.length != 2, This throws out all times that there is any combination of spaces and Letters that doesn't result in a 2 d array. So in your example of "Name: ___" (underscores are spaces), the split(' ') function will auto that to be a 1-length array containing just "Name:", in which case it would have thrown eariler when it checked for .length != 2. This of course only works assuming i implement a better validation of header like .startswith() that you mentioned earlier.

raymondshiner: Cause this validation doesn't account for "Name:Raymond Shiner" because of lackluster checking before

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
KennethWhite: In hindsight I have no idea why we don't have even basic checks against certain things, like `User` being null.

SecretSanta/src/SecretSanta.Library/FileParser.cs
KennethWhite: I agree with this as I am making some assumptions about the state of the `StreamReader` when it returns. It does seem better to avoid the issue altogether by only doling out strings to the parsing methods.

SecretSanta/src/SecretSanta.Library/FileParser.cs
KennethWhite: For the case with no spaces, I think we made some assumptions regarding the format of the file and that there would be at least one space. As for the case with two or more spaces, looking at the [MSDN for String.Split()](https://docs.microsoft.com/en-us/dotnet/api/system.string.split?view=netcore-2.2) there is an [Enum for StringSplitOptions](https://docs.microsoft.com/en-us/dotnet/api/system.stringsplitoptions?redirectedfrom=MSDN&view=netcore-2.2) which has a particular flag `StringSplitOptions.RemoveEmptyEntries` which would have been useful as an optional second parameter to `String.Split()`

SecretSanta/src/SecretSanta.Library/FileParser.cs
KennethWhite: We definitely dropped the ball with some basic validation for the properties.

MarkMichaelis:
One more thing to note, you don't seem to have ever actually imported the data to the database?  You parsed the file but never imported the data?

PR #49, opened 2019-01-23
Assignment2

COsborn2:


SecretSanta/src/SecretSanta.Import/ImportUtility.cs
```diff
@@ -0,0 +1,67 @@
...
+        private const string ProperFormatException =
+            "Format must be either \"Name: <Firstname> <Lastname>\" or \"Name: <Lastname>, <Firstname>\"";
+        
+        public static User Import(string fileName)
+        {
+            if(String.IsNullOrEmpty(fileName)) { throw new NullReferenceException(); }
+            
+            string headerLine = GetHeader(fileName);
+            
+            Console.WriteLine(headerLine);
```
efueng: Not sure why you're printing here. If it's for testing just remember to get rid of it for the final PR.

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            if (!headerLine.StartsWith("Name:"))
+            {
+                throw new ArgumentException("Header must start with \"Name:\"", headerLine);
+            }
+            
+            headerLine = headerLine.Replace("Name: ", "")
```
efueng: This is cool, I might have to steal this lol

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
```diff
@@ -0,0 +1,67 @@
...
+            if (headerLine.Contains(','))
+            {
+                splitName = headerLine.Split(',').Reverse().ToList();
+            }
+            else
+            {
+                splitName = headerLine.Split().ToList();
+            }
+
+            if (splitName.Count != 2)
```
efueng: Something to consider is if you're going to be supporting middle names in the future

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                throw new ArgumentException(ProperFormatException, headerLine);
+            }
+
+            string firstName = splitName[0].Trim();
+            string lastName = splitName[1].Trim();
+
+            if (String.IsNullOrEmpty(firstName) || String.IsNullOrEmpty(lastName))
```
efueng: You might consider moving name validation to properties in the User class

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
```diff
@@ -0,0 +1,67 @@
...
+            {
+                throw new ArgumentException(ProperFormatException, headerLine);
+            }
+            
+            return new User()
+            {
+                FirstName = firstName,
+                LastName = lastName
+            };
```
efueng: Consider splitting the Import method into smaller methods. It would probably make it easier to test, as well.

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        private static string GetHeader(string fileName)
```
efueng: This might just be a style thing but since your method is called GetHeader, consider moving header validation in here. Right now GetHeader really just returns the first line of a file whether or not it's a valid header.

SecretSanta/test/SecretSanta.Import.Tests/ImportUtilityTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            return tempFileName;
+        }
+        
+        [DataTestMethod]
+        [DataRow("Name: Inigo Montoya", "Inigo", "Montoya")]
+        [DataRow("Name: Princess Buttercup", "Princess", "Buttercup")]
+        [DataRow("Name: Person McName", "Person", "McName")]
+        [DataRow("Name: Person Hyphen-Name", "Person", "Hyphen-Name")]
```
efueng: Good idea on testing with hyphens; I didn't think about that

SecretSanta/test/SecretSanta.Import.Tests/ImportUtilityTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        [DataRow("Name: McName, Person", "Person", "McName")]
+        [DataRow("Name: Hyphen-Name, Person", "Person", "Hyphen-Name")]
+        public void Import_LastnameCommaFirstname_ProperFormat_Success(string toWrite, string firstName, string lastName)
+        {
+            var tempFileName = WriteTemporaryFile(new List<string>(){ toWrite });
+
+            User user = ImportUtility.Import(tempFileName);
+            
+            Assert.AreEqual(firstName, user.FirstName);
+            Assert.AreEqual(lastName, user.LastName);
```
efueng: I like the way your tests are set up, but you're only testing with good data. You probably already know that though.

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
COsborn2: Good catch. I guess I forgot to take that before I pushed.

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
COsborn2: Thats a good idea, I'll do that.

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            var headerLine = File.ReadLines(fileName).First().Trim();
+
```
raymondshiner: Does this implementation account for both an absolute and a relative path for the filename? If so how?

COsborn2: It does. Since I'm just calling the built in C# IO Methods it is handled automatically. More on that [here](https://docs.microsoft.com/en-us/dotnet/api/system.io.file?view=netframework-4.7.2). About halfway down the page it shows some examples of valid paths. And [here](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.readlines?view=netframework-4.7.2) you can see more on the File.ReadLines method I used. Hopefully that helps. I think Mark wanted to make sure our tests used relative paths so we dont reference a path that doesn't exist on another machine. If we used absolute paths for that, then our tests wouldn't be portable, which is necessary for Azure to run them. Hope this helps!

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
MarkMichaelis: Yes @efueng, good feedback.

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
```diff
@@ -0,0 +1,77 @@
...
+            return toReturn;
+        }
+
+        private static string GetHeader(string properlyQualifiedFilename)
+        {
+            var headerLine = File.ReadLines(properlyQualifiedFilename).First().Trim();
+
+            return headerLine.StartsWith("Name:")
+                ? headerLine
+                : throw new ArgumentException("Header must start with \"Name:\"", nameof(headerLine));
```
MarkMichaelis: No const this time... how come?  Consistency is preferable IMO even if you don't go with const either time. :)

SecretSanta/src/SecretSanta.Import/ImportUtility.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        private static List<Gift> GetGifts(string fileName, User user)
+        {
+            return File.ReadLines(fileName)
+                .Skip(1)
+                .Select(line => line.Trim())
+                .Where(line => !string.IsNullOrEmpty(line))
```
MarkMichaelis: Validation logic is preferable in the `Gift` class.  You could even have a static verify method or `TryParse()` method.

SecretSanta/test/SecretSanta.Import.Tests/ImportUtilityTests.cs
```diff
@@ -0,0 +1,183 @@
...
+        private string WriteTemporaryFile(IEnumerable<string> toWrite)
+        {
+            var tempFileName = Path.GetTempFileName();
+
+            using (var streamWriter = new StreamWriter(tempFileName))
+            {
+                foreach (var cur in toWrite)
+                    streamWriter.WriteLine(cur);
+
+                streamWriter.Close();
```
MarkMichaelis: I believe that if you use a using statment you don't need to call `Close()`.

SecretSanta/test/SecretSanta.Import.Tests/ImportUtilityTests.cs
```diff
@@ -0,0 +1,183 @@
...
+            }
+
+            CreatedFiles.Add(tempFileName);
+
+            return tempFileName;
+        }
+
+        private string WriteFileAtPath(string path, IEnumerable<string> toWrite)
+        {
+            for (var i = 0; i < 100; i++) // attempt to create random file 100 times before giving up
```
MarkMichaelis: It is a pain that you have to do this but you handled it well. 
After checking the [source code](https://referencesource.microsoft.com/#mscorlib/system/io/path.cs,efb113f637a6bb47), I'd be tempted just to use GetTempFileName() unless I really needed it in my own directory. Perhaps writing my own library using your logic that wrapped GetRandomFileName() is the best.
That said, this is a test after all.  It probably isn't worth the effort for what is extremely unlikely to occur.

SecretSanta/test/SecretSanta.Import.Tests/ImportUtilityTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                }
+
+                CreatedFiles.Add(newFileName);
+                return newFileName;
+            }
+
+            throw new IOException("Random file could not be created");
+        }
+
+        [DataTestMethod]
```
MarkMichaelis: Great selection of test data.

SecretSanta/test/SecretSanta.Import.Tests/ImportUtilityTests.cs
```diff
@@ -0,0 +1,183 @@
...
+        [DataRow("Name: Inigo Montoya", "", "Inigo", "Montoya")]
+        public void Import_WithFullGiftList_ProperFormat_Success(string headerLine, string giftsToUse, string firstName,
+            string lastName)
+        {
+            var testUser = new User
+            {
+                FirstName = firstName,
+                LastName = lastName
+            };
+            var gifts = giftsToUse
```
MarkMichaelis: Nice use of LINQ.

SecretSanta/test/SecretSanta.Import.Tests/ImportUtilityTests.cs
```diff
@@ -0,0 +1,183 @@
...
+                .Split(',')
+                .Select(line => line.Trim())
+                .Where(line => line != "")
+                .Select(line => new Gift
+                {
+                    Title = line,
+                    User = testUser
+                })
+                .ToList();
```
MarkMichaelis: Generally... avoid this conversion unless you have a specific reason to need `List<T>`.

SecretSanta/test/SecretSanta.Import.Tests/ImportUtilityTests.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        [DataTestMethod]
+        [ExpectedException(typeof(NullReferenceException))]
+        public void Import_FilenameNull_NullReferenceException()
+        {
+            ImportUtility.Import(null);
+        }
+
+        [DataTestMethod]
```
MarkMichaelis: Again... great selection of test data.

MarkMichaelis:

PR #50, opened 2019-01-23
Assignment2_Import

KyleBurgi:


SecretSanta/SecretSanta.Import.Tests/ImportTests.cs
```diff
@@ -0,0 +1,51 @@
...
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            if (File.Exists("tempListOne.txt")) File.Delete("tempListOne.txt");
+            if (File.Exists("tempListTwo.txt")) File.Delete("tempListTwo.txt");
+
+            string[] testListOne = { "Name:Kyle Burgi", "Pokemon Cards", "Shiny Computer", "Millions of Dollars" };
+            string[] testListTwo = { "Name:Burgi, Kyle", "Digimon Cards", "Shiny Computer", "Millions of Dollars", "A Pony." };
+
+            File.WriteAllLines("tmp.txt", testListOne);
```
KennethWhite: Unless I am mistaken, in your initialize method you clean up two files `tempListOne.txt` and `tempListTwo.txt` but are creating files actually names `tmp.txt` and `tmp2.txt`

SecretSanta/SecretSanta.Import/Import.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            if (regexArray[0].Equals("Name") == false)
+                return false;
+
+            return true;
+        }
+
+        public User ReadHeader()
+        {
+            string header = File.ReadLines(DirectoryPath).First();
```
KennethWhite: I think this will miss the edge case where there is one or more blank lines at the beginning of a file, before the header line.

SecretSanta/SecretSanta.Import/Import.cs
```diff
@@ -0,0 +1,82 @@
...
+                return new User(nameArray[1], nameArray[0]);
+            }
+
+            if (nameArray.Length == 1)
+            {
+                string[] normalNameSplit = nameArray[0].Split(' ');
+                return new User(normalNameSplit[0], normalNameSplit[1]);
+            }
+
+            throw new ArgumentException(nameof(DirectoryPath));
```
KennethWhite: This might be a style choice, but this method is a little long and checks a lot of things. It might be worth breaking it up into a couple of methods.

SecretSanta/SecretSanta.Import/Import.cs
```diff
@@ -0,0 +1,82 @@
...
+        public List<Gift> ReadFileGifts(User user)
+        {
+            if (user == null)
+                throw new ArgumentNullException(nameof(user));
+
+            List<Gift> giftList = new List<Gift>();
+
+            File.ReadLines(DirectoryPath).Skip(1);
+
+            return null;
```
KennethWhite: I dont think you are using this anywhere, but you should probably replace the return statement.
```suggestion
            return giftList;
```

SecretSanta/SecretSanta.Import/Import.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        public List<Gift> ReadFileGifts(User user)
+        {
+            if (user == null)
+                throw new ArgumentNullException(nameof(user));
+
+            List<Gift> giftList = new List<Gift>();
+            int count = 0;
+            string giftString = null;
```
MarkMichaelis: Minor: I would try avoid using data type names in variable names.  Consider using "Text" rather than "String".

SecretSanta/SecretSanta.Import/Import.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            var fileStream = new FileStream(DirectoryPath, FileMode.Open, FileAccess.Read);
+            using (var streamReader = new StreamReader(fileStream, Encoding.UTF8))
+            {
+                while ((giftString = streamReader.ReadLine()) != null)
+                {
+                    if (count == 0) { }
+                    else
+                        giftList.Add(new Gift(giftString, count));
```
MarkMichaelis: There is no validation on the values used when instantiating a Gift.  Thus null, empty string would be allowable.

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -0,0 +1,22 @@
+﻿using SecretSanta.Domain.Models;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain
+{
+    public class Gift : Entity
+    {
+        public string GiftTitle { get; set; }
```
MarkMichaelis: I would expect validation on these properties.

SecretSanta/SecretSanta.Import.Tests/ImportTests.cs
```diff
@@ -0,0 +1,138 @@
...
+using System.Text;
+using SecretSanta.Import;
+using System.IO;
+using SecretSanta.Domain;
+using System.Linq;
+
+namespace SecretSanta.Import.Tests
+{
+    [TestClass]
+    public class ImportTests
```
MarkMichaelis: There is a significant amount of refactoring possible here that would clean up the code significantly.

SecretSanta/SecretSanta.Import.Tests/ImportTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        [DataRow("/goodTestListTwo.txt")]
+        public void CheckHeader_Name_Success(string path)
+        {
+            Import import = new Import();
+            Assert.IsTrue(import.ProperHeaderExists(Path.GetTempPath() + path));
+
+        }
+
+        [TestMethod]
+        [DataRow("/goodTestListOne.txt", "Kyle", "Burgi")]
```
MarkMichaelis: Great use of DataRowAttribute.  Good job with that.

SecretSanta/SecretSanta.Import.Tests/ImportTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        [TestMethod]
+        [DataRow("/goodTestListEmpty.txt")]
+        [DataRow("/goodTestListEmptyTwo.txt")]
+        [ExpectedException(typeof(ArgumentException))]
+        public void CheckWishlist_IsEmpty_ThrowArgument(string path)
+        {
+            Import import = new Import(Path.GetTempPath() + path);
+            User checkUser = import.ReadHeader();
+            import.ReadFileGifts(checkUser);
+            checkUser.PrintWishList().Contains("Millions of Dollars");
```
MarkMichaelis: What is this line for?  Don't you expect an exception to be thrown before this?

SecretSanta/SecretSanta.Import.Tests/ImportTests.cs
```diff
@@ -0,0 +1,138 @@
...
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Import_filenameIsNull_ThrowsException()
+        {
+            Import import = new Import(null);
+        }
+
+
+        [TestMethod]
+        [DataRow("/ThisIsABadPath.txt")]
```
MarkMichaelis: I'm not quite understanding what is being tested for in the DataRow attributes 2-4?  It seems like they are testing the same thing.

MarkMichaelis:

PR #54, opened 2019-01-29
Assignment3

grantdwoods:


SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -19,47 +32,167 @@ public void GiftController_RequiresGiftService()
...
-            Assert.AreEqual(gift.Url, resultGift.Url);
-            Assert.AreEqual(gift.OrderOfImportance, resultGift.OrderOfImportance);
+
+            Assert.AreEqual<int>(gift.UserId, resultGift.UserId);
+            Assert.AreEqual<int>(gift.Id, resultGift.Id);
+            Assert.AreEqual<string>(gift.Title, resultGift.Title);
+            Assert.AreEqual<string>(gift.Description, resultGift.Description);
+            Assert.AreEqual<string>(gift.Url, resultGift.Url);
+            Assert.AreEqual<int>(gift.OrderOfImportance, resultGift.OrderOfImportance);
+            MockGiftService.VerifyAll();
```
KennethWhite: I didnt use Moq myself, but your tests here make me wish I had. Pretty slick implementation here.

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -19,47 +32,167 @@ public void GiftController_RequiresGiftService()
...
+
+            MockGiftService.Setup(x => x.AddGiftToUser(It.IsAny<int>(), It.IsAny<Gift>()))
+                .Callback((int userid, Gift giftIn) => { giftIn.UserId = userid; });
+
+            var controller = new GiftController(MockGiftService.Object);
+            var result = (CreatedResult)controller.PostGiftToUser(2, gift);
+            
+            var returendGift = (DTO.Gift)result.Value;
+
+            Assert.AreEqual<int?>(201, result.StatusCode);
```
KennethWhite: I didnt think to verify the status codes in my tests, nice and thorough checking here.

grantdwoods:
@Keboo 
As I began working on Assignment 4 I came to the realization that I forgot to implement and test removing users from a group.
Just making sure to mention it because I'm not trying to sneak one by, I genuinely thought the assignment was complete.

SecretSanta/test/SecretSanta.Api.Tests/GroupControllerTests.cs
```diff
@@ -0,0 +1,191 @@
...
+        [DataRow(0)]
+        [DataRow(-1)]
+        public void GetUsersInGroup_GroupIdLessthanOne_Return404(int groupId)
+        {
+            var controller = new GroupController(MockGroupService.Object);
+
+            var result = (NotFoundResult)controller.GetUsersInGroup(groupId);
+
+            Assert.AreEqual<int?>(404, result.StatusCode);
+            MockGroupService.Verify(x => x.GetUsers(It.IsAny<int>()), Times.Never);
```
Keboo: There is a way to reduce these unit tests further when using Moq. By default the mock object are created "loose" meaning that if it encounters a method without a corresponding `Setup` it will simply return the default return value. However, you can create the mock objects as "strict" meaning that it will throw an exception for any call that does not contain a `Setup`. Hence you can eliminate the Verify calls that look for `Times.Never`.

SecretSanta/test/SecretSanta.Api.Tests/UserControllerTests.cs
```diff
@@ -0,0 +1,109 @@
...
+            var result = (CreatedResult)controller.PostUser(user);
+
+            var returnedUser = (DTO.User)result.Value;
+
+            Assert.AreEqual<int?>(201, result.StatusCode);
+            Assert.AreNotEqual<int>(0, returnedUser.Id);
+            Assert.AreEqual<string>($"api/User/{returnedUser.Id}", result.Location);
+            Assert.AreEqual<string>("Grant", returnedUser.FirstName);
+            Assert.AreEqual<string>("Woods", returnedUser.LastName);
+            MockUserService.VerifyAll();
```
Keboo: Typically with AutoMocker you would let the AutoMocker instance verify all expectations on every mock that it is tracking. 
```suggestion
            Mocker.VerifyAll();
```

PR #56, opened 2019-01-30
Assignment3

RichardTeller:
Still need to implement Controller classes and tests.

DTO objects and services have been created/updated.


SecretSanta/src/SecretSanta.Api/DTO/Gift.cs
```diff
@@ -9,13 +9,14 @@ public class Gift
         public string Description { get; set; }
         public int OrderOfImportance { get; set; }
         public string Url { get; set; }
+        public int UserId { get; set; }
```
Cwhite534: I am not sure if the userId should be exposed. I didn't expose Entity relationships in any of the DTO models.

SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
```diff
@@ -20,15 +20,31 @@ public GiftController(IGiftService giftService)
...
             }
             List<Gift> databaseUsers = _GiftService.GetGiftsForUser(userId);
-
+            
             return databaseUsers.Select(x => new DTO.Gift(x)).ToList();
         }
+
+        // POST api/Gift/4
+        [HttpPost("{userId}")]
+        public ActionResult AddGiftToUser(DTO.Gift gift, int userId)
```
Sturminator: I don't know if you know this, but you specify a type for ActionResult like in the GetGiftsForUser function. In this case you could specify ActionResult<DTO.Gift> and return a new DTO.Gift. I think this could used in cases of displaying more detailed information on what has happened to the user.

SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
```diff
@@ -20,15 +20,31 @@ public GiftController(IGiftService giftService)
...
+        {
+            if (userId <= 0)
+            {
+                return NotFound();
+            }
+            if (gift == null)
+            {
+                return BadRequest();
+            }
+            _GiftService.AddGiftToUser(userId, null);
```
Sturminator: I'm pretty sure this is unfinished, yes? I saw you've begun creating your ToDomain functions and I'm guessing this is where you will plan to use it. Other than that, this looks good.

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -28,9 +29,47 @@ public Group UpdateGroup(Group @group)
...
+            return user;
+        }
+
+        public User RemoveUserFromGroup(int groupId, User user)
+        {
+            if (user == null) throw new ArgumentNullException(nameof(user));
+
+            Group group = DbContext.Groups.Include(g => g.GroupUsers).SingleOrDefault(g => g.Id == groupId);
+            GroupUser groupUser = group?.GroupUsers.Single(gu => gu.GroupId == groupId && gu.UserId == user.Id);
+            group?.GroupUsers.Remove(groupUser);
```
Keboo: This doesn't work (a unit test would have caught this). This is trying to remove an object from the `GroupUsers` list that is the same instance as `groupUser`, but since this is a new instances that you just created it will never be in that list. What you actually wanted to do was search the list and see if there is an item in the list with the passed in `groupId` and `userId` and remove that.

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -17,49 +18,163 @@ public void GiftController_RequiresGiftService()
...
+            var controller = new GiftController(testService);
+
+            ActionResult<List<DTO.Gift>> result = controller.QueryAllGiftsForUser(2);
+
+            Assert.AreEqual(2, testService.LastModifiedUserId);
+            DTO.Gift resultGift = result.Value.Single();
+            Assert.AreEqual(gift.Id, resultGift.Id);
+        }
+
+        private Gift CreateGiftWithId(int id)
```
Keboo: :+1:

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -17,49 +18,163 @@ public void GiftController_RequiresGiftService()
...
+        }
+        
+        [TestMethod]
+        public void RemoveGiftForUser_GiftIsRemoved()
+        {
+            var testService = new TestableGiftService();
+            var controller = new GiftController(testService);
+
+            var giftToRemove = CreateGiftWithId(2);
+            controller.RemoveGift(new DTO.Gift(giftToRemove));
```
Keboo: Would have been nice to check the return from `RemoveGift` as well.

SecretSanta/test/SecretSanta.Api.Tests/TestableGiftService.cs
```diff
@@ -1,18 +1,39 @@
...
 
 namespace SecretSanta.Api.Tests
 {
     public class TestableGiftService : IGiftService
     {
-        public List<Gift> ToReturn { get; set; }
-        public int GetGiftsForUser_UserId { get; set; }
+        public List<Gift> GiftsToReturn { get; set; }
+        public int LastModifiedUserId { get; set; }
+        public Gift LastModifiedGift { get; set; }
```
Keboo: The problem with re-using the same properties for each method is you cannot actually verify that any given method was the one invoked. This would allow a method to invoke the wrong service method and still have your unit test pass.
You want to have separate properties for each method.

PR #57, opened 2019-01-30
Assignment3

tdang2018:
Still need to implement UserController and UserControllerTests, GroupController and GroupControllerTests.

### DIFF
--- a/SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
@@ -27,8 +27,28 @@ namespace SecretSanta.Api.Controllers
                 return NotFound();
             }
             List<Gift> databaseUsers = _GiftService.GetGiftsForUser(userId);
-
+            
             return databaseUsers.Select(x => new DTO.Gift(x)).ToList();
+        }
+
+        //POST api/Gift/4
+        [HttpPost("{userId}")]
+        public ActionResult AddGiftToUser(DTO.Gift gift, int userId)
+        {
+            if (userId <= 0)
+            {
+                return NotFound();
+            }
+
+            if (gift == null)
+            {
+                return BadRequest();
+            }
+
+            _GiftService.AddGiftToUser(userId, DTO.Gift.ToEntity(gift));
+            return Ok();
+
+            //return databaseUsers.Select(x => new DTO.Gift(x)).ToList();
         }
     }
 }
--- a/SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
@@ -51,7 +51,7 @@ namespace SecretSanta.Api.Controllers
 
 
         // PUT api/values/5
-        [HttpPut("{id}")]
+        [HttpPut("{userId}")]
         public ActionResult UpdateGiftToUser(DTO.Gift gift, int userId)
         {
             if (userId <= 0)
@@ -69,7 +69,7 @@ namespace SecretSanta.Api.Controllers
             return Ok();
         }
 
-        [HttpDelete("{id}")]
+        [HttpDelete]
         public ActionResult RemoveGift(DTO.Gift gift)
         {
             if (gift == null)
--- a/SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
@@ -46,9 +46,42 @@ namespace SecretSanta.Api.Controllers
             }
 
             _GiftService.AddGiftToUser(userId, DTO.Gift.ToEntity(gift));
-            return Ok();
-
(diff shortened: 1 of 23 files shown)


SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
```diff
@@ -27,8 +27,61 @@ public ActionResult<List<DTO.Gift>> GetGiftForUser(int userId)
...
+                return BadRequest();
+            }
+
+            _GiftService.AddGiftToUser(userId, DTO.Gift.ToEntity(gift));
+            return Ok();            
+        }
+
+
+        // PUT api/values/5
+        [HttpPut("{id}")]
```
efueng: You should be consistent with names of the string params for your Http attributes. For example, above you use `[HttpPost("{userId}")]` and here you use `[HttpPost("{id}")]`.

SecretSanta/src/SecretSanta.Api/DTO/Gift.cs
```diff
@@ -25,5 +25,21 @@ public Gift(SecretSanta.Domain.Models.Gift gift)
             OrderOfImportance = gift.OrderOfImportance;
             Url = gift.Url;
         }
+
+        public static SecretSanta.Domain.Models.Gift ToEntity(DTO.Gift gift)
```
efueng: I had a long internal debate about this. I implemented this in the controller classes since they work with both the DTO objects and the Model/Entity objects. Doing this in the DTO classes results in cleaner code in the controllers, but now the DTOs have to "know" about their Domain/Entity counterparts. In the end I guess it's just a style/implementation tradeoff.

SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
```diff
@@ -27,8 +27,61 @@ public ActionResult<List<DTO.Gift>> GetGiftForUser(int userId)
...
+            }
+
+            _GiftService.AddGiftToUser(userId, DTO.Gift.ToEntity(gift));
+            return Ok();            
+        }
+
+
+        // PUT api/values/5
+        [HttpPut("{userId}")]
+        public ActionResult UpdateGiftToUser(DTO.Gift gift, int userId)
```
Keboo: Minor issue here. Your controller is named "UpdateGiftToUser" while the service method is "UpdateGiftForUser". I would suggest that this method should also be named UpdateGiftForUser

SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
```diff
@@ -0,0 +1,99 @@
...
+                return BadRequest();
+            return databaseGroups.Select(x => new DTO.Group(x)).ToList();
+        }
+
+        [HttpGet("{groupId}")]
+        public ActionResult<List<DTO.User>> FetchAllUsersInGroup(int groupId)
+        {
+            if (groupId <= 0) return NotFound();
+
+            return _GroupService.FetchAllUsersInGroup(groupId).Select(user => new DTO.User(user)).ToList();
```
Keboo: Your service can return `null` from `FetchAllUsersInGroup` you should handle that case here.

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -14,23 +15,67 @@ public GroupService(ApplicationDbContext dbContext)
...
+            return group?.GroupUsers.Select(gu => gu.User).ToList();
+        }
+
+        public User AddUserToGroup(int groupId, User user)
+        {
+            if (user == null) throw new ArgumentNullException(nameof(user));
+
+            var group = DbContext.Groups.Include(g => g.GroupUsers)
+                .SingleOrDefault(g => g.Id == groupId);
+            var groupUser = new GroupUser { GroupId = groupId, UserId = user.Id, User = user };
```
Keboo: You technically don't need to set the `User` property here; just the IDs are enough.

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -14,23 +15,67 @@ public GroupService(ApplicationDbContext dbContext)
...
+        }
+
+        public User RemoveUserFromGroup(int groupId, User user)
+        {
+            if (user == null) throw new ArgumentNullException(nameof(user));
+
+            var group = DbContext.Groups.Include(g => g.GroupUsers)
+                .SingleOrDefault(g => g.Id == groupId);
+            var userGroupToRemove = group?.GroupUsers.Single(ug => ug.GroupId == groupId && ug.UserId == user.Id);
+            group?.GroupUsers.Remove(userGroupToRemove);
```
Keboo: This doesn't work (a unit test would have caught this). This is trying to remove an object from the `GroupUsers` list that is the same instance as `userGroupToRemove`, but since this is a new instances that you just created it will never be in that list. What you actually wanted to do was search the list and see if there is an item in the list with the passed in `groupId` and `userId` and remove that.

PR #58, opened 2019-01-30
Assignment 3 - Brad Howard

BHoward85:
UpdateUpdate: I merge with the remote to clear the conflict.


SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -0,0 +1,91 @@
...
+        public ActionResult<DTO.User> FindUser(int id)
+        {
+            if(id <= 0)
+            {
+                return NotFound();
+            }
+
+            DTO.User user = new DTO.User(_UserService.Find(id));
+
+            return user;
```
KennethWhite: This is a styling comment, but you could reduce this to a single line
```suggestion
            return new DTO.User(_UserService.Find(id));
```

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                return NotFound();
+            }
+
+            if(upDatedUser == null)
+            {
+                return BadRequest();
+            }
+
+            User user = new User()
```
KennethWhite: Same comment as above with groups, this feels like an odd place for this.

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -1,52 +1,130 @@
...
         {
-            DbContext = dbContext ?? throw new ArgumentNullException(nameof(dbContext));
+            DbContext = context;
         }
 
-        public Gift AddGiftToUser(int userId, Gift gift)
+        public bool CreateGift(User user, Gift gift)
         {
-            if (gift == null) throw new ArgumentNullException(nameof(gift));
+            bool test = false;
```
KennethWhite: A little pedantic here, but test is not a very descriptive variable name

SecretSanta/src/SecretSanta.Domain/Models/User.cs
```diff
@@ -1,13 +1,20 @@
...
     {
-        public int Id { get; set; }
-        public string FirstName { get; set; }
-        public string LastName { get; set; }
-        public List<Gift> Gifts { get; set; }
-        public List<GroupUser> GroupUsers { get; set; }
+        public string First { get; set; }
+        public string Last { get; set; }
+        public ICollection<UserGroup> UserGroups { get; set; }
+        public ICollection<Gift> Gifts { get; set; }
```
KennethWhite: Out of curiosity, why did you change from a `List` to an `ICollection` if in your implementation you end up casting them to `List<Gift>`?

SecretSanta/src/SecretSanta.Domain/Services/GiftService.cs
```diff
@@ -1,52 +1,130 @@
...
+            if(id > 0)
+            {
+                foreach(Gift g in giftList)
+                {
+                    if (g.Id == id)
+                        gift = g;
+                }
+            }
+
             return gift;
```
KennethWhite: If the ID provided is 0 or another invalid value, this method will return a `null`, which is not something you are checking for in your controller.

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -1,36 +1,186 @@
...
+
+            return true;
+        }
+
+        public bool AddUser(User user, int id)
+        {
+            Group group = FindGroup(id);
+            UserGroup userGroup = DbContext.UserGroups.Find(user.Id, id);
+
+            if (group != null)
```
KennethWhite: There is a [Null Conditional Operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-conditional-operators) that might be useful to make these null checks more compact.

SecretSanta/test/SecretSanta.Api.Tests/GroupControllerTests.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            TestableGroupService service = new TestableGroupService()
+            {
+                ToReturnGetAllGroup = new List<Group>()
+            };
+            GroupController controller = new GroupController(service);
+
+            Group group = new Group() { Title = "Battleship Battle Group 1" };
+
+            service.ToReturnGetAllGroup.Add(group);
```
KennethWhite: This name was very confusing, I had to go look at the implementation.

PR #60, opened 2019-01-30
Assignment3

KennethWhite:
Collaboration with @Cwhite534 and @KennethWhite


SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -28,22 +28,25 @@ public Group UpdateGroup(Group group)
...
+            DbContext.SaveChanges();
         }
 
-        public Group AddUserToGroup(int userId, Group @group)
+        public Group AddUserToGroup(int userId, Group group)
         {
             throw new NotImplementedException();
         }
 
-        public void RemoveUserFromGroup(int userId, Group @group)
```
tdang2018: Hi Kevin. Why do you replace Group @group to Group group?

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
```diff
@@ -1,7 +1,56 @@
...
+            var groupToUpdate = Groups.Single(g => g.Id == group.Id);
+            groupToUpdate.Name = group.Name;
+            groupToUpdate.GroupUsers = group.GroupUsers;
+            return group;
+        }
+
+        public void RemoveGroup(Group group)
+        {
+            LastGroupModified = group.Id;
+            Groups.Remove(Groups.Single(g => g.Id == group.Id));
```
tdang2018: So you are using lamda expression in this line. 
This line is trying to delete the group where g.Id = group.Id. ? My question is what does Groups.Single do?

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
```diff
@@ -1,7 +1,56 @@
...
+
+        public List<User> FetchAllUsersInGroup(Group group)
+        {
+            throw new NotImplementedException();
+        }
+
+        public List<Group> FetchAllGroups()
+        {
+            throw new NotImplementedException();
+        }
```
tdang2018: You forgot to implement the above methods : 

- AddUserToGroup(int userId, Group group)
-  RemoveUserFromGroup(int userId, Group group)
- List<Group> FetchAllGroups()
-

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
```diff
@@ -1,7 +1,56 @@
...
+using SecretSanta.Domain.Models;
+using SecretSanta.Domain.Services;
+
 namespace SecretSanta.Api.Tests
 {
-    public class TestableGroupService
+    public class TestableGroupService : IGroupService
     {
+        public List<Group> Groups { get; set; }
+        public int LastGroupModified;
```
tdang2018: What does property LastGroupModified do? and when do you use it?

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
```diff
@@ -1,7 +1,56 @@
...
+        public Group AddGroup(Group group)
+        {
+            LastGroupModified = group.Id;
+            Groups.Add(group);
+            return group;
+        }
+
+        public Group UpdateGroup(Group group)
+        {
+            LastGroupModified = group.Id;
```
tdang2018: Why do you have to use LastGroupModifiedin this line? I dont understand. Can you explain for me please?

SecretSanta/src/SecretSanta.Api/DTO/Group.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+
+namespace SecretSanta.Api.DTO
+{
+    public class Group
+    {
+        public int Id { get; set; }
+        public string Name { get; set; }
+
```
tdang2018: I think you should implement default constructor public Group() { }  like you did on the other classes.

SecretSanta/src/SecretSanta.Api/Startup.cs
```diff
@@ -38,6 +38,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddSwaggerGen(c =>
             {
```
tdang2018: What does line code mean?  
 c.SwaggerDoc("v1", new Info { Title = "My API", Version = "v1" });
                c.CustomSchemaIds(x => x.FullName);

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -19,47 +19,161 @@ public void GiftController_RequiresGiftService()
...
+        [TestMethod]
+        public void RemoveGiftForUser_RequiresGift()
+        {
+            var testService = new TestableGiftService();
+            var controller = new GiftController(testService);
+            ActionResult<DTO.Gift> result = controller.RemoveGift(null);
+
+            Assert.IsTrue(result.Result is BadRequestResult);
+            //This check ensures that the service was not called
+            Assert.IsNull(testService.LastModifiedGift);
```
tdang2018: What does line do ? 
Assert.IsNull(testService.LastModifiedGift); Why do you have to check LastModiefiedGift? and how to use it?

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
Keboo: @tdang2018 this escapes the C# keyword "group" so it can be used as a parameter name. See the [docs](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/verbatim) or page 15 in the book.

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
Cwhite534: @tdang2018 In this method I kind of got away from myself, and added functionality It didn't need. It looks as if you were looking at an older commit as it was changed. But anyway `Groups.Single(g => g.Id == group.Id)` will return the group in the list with the correct Id, and then it gets removed. Im not entirely sure why I didn't just call remove on the param group.

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
Cwhite534: This was an older commit.

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
Cwhite534: The LastGroupModified is assigned whenever the controller method gets sent a Group object that is then passed to the service. It allows me to make sure that the DTO.Group object passed to the controller method is correctly converted to a Domain.Models.Group and passed to the service method.

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
Cwhite534: As said above this is for testing the functionality of the GroupControllers UpdateGroup method. This again has been updated to ` LastGroupModified = group;`

SecretSanta/src/SecretSanta.Api/DTO/Group.cs
Cwhite534: Good catch

SecretSanta/src/SecretSanta.Api/Startup.cs
Cwhite534: The first line is the default registration code for swagger. The second tells swagger to use full class names like `DTO.Group` instead of just `Group` I did this to correct an issue I was having, but the issue was caused because I was passing a `Domain.Model.Group` into the controller method instead of 'DTO.Group'. This line will be removed

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
tdang2018: @Cwhite534 : Thanks Casey.

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
tdang2018: @Cwhite534 : Now it makes sense to me now. Thanks Casey.

SecretSanta/test/SecretSanta.Api.Tests/TestableGiftService.cs
```diff
@@ -1,18 +1,40 @@
...
 
 namespace SecretSanta.Api.Tests
 {
     public class TestableGiftService : IGiftService
     {
-        public List<Gift> ToReturn { get; set; }
-        public int GetGiftsForUser_UserId { get; set; }
+        public List<Gift> GiftsToReturn { get; set; }
+        public int LastModifiedUserId { get; set; }
+        public Gift LastModifiedGift { get; set; }
```
Keboo: For these testable objects you want to use separate properties for each method. By re-using the same properties in each method it means that you have no way of telling which method was actually used.

PR #62, opened 2019-01-30
Assignment3

COsborn2:


SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -0,0 +1,50 @@
...
+    public class UserController : ControllerBase
+    {
+        private readonly IUserService _userService;
+
+        public UserController(IUserService userService)
+        {
+            _userService = userService ?? throw new ArgumentNullException(nameof(userService));
+        }
+
+        // GET api/Gift/5
```
raymondshiner: Here you are using a HttpGet for adding something to the database. you should be using HttpPost instead and seperatley implement a GetUser method seperately that grabs a user from the DB. I'm like 90% sure we are supposed to also do gets for everything.

SecretSanta/test/SecretSanta.Api.Tests/UserControllerTests.cs
```diff
@@ -0,0 +1,121 @@
+using System;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SecretSanta.Api.Controllers;
+using SecretSanta.Api.DTO;
+
```
raymondshiner: I would consider refactoring your implmentation of these differnet User classes to be that  you always include Domain.Models in your using statements and you always use DTO.User explicitly when you need it. Otherwise you have half of your code calling "User" objects and another half of your code over in the Actual DB Services also calling "User" objects, but they are both calling different things. I think reworking it would make it easier to understand longterm.

raymondshiner: This would be like all over your code including your controllers

SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
```diff
@@ -26,9 +26,60 @@ public ActionResult<List<DTO.Gift>> GetGiftForUser(int userId)
...
+            {
+                return NotFound(gift);
+            }
+
+            if (gift == null)
+            {
+                return BadRequest();
+            }
+
+            Gift addedGift = _giftService.AddGiftToUser(userId, DTO.Gift.ToEntity(gift));
```
joeriddles: Good use of a static method here when creating a Model.Gift from a DTO.Gift

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
joeriddles: I believe @raymondshiner is correct in that is should be `[HttpPost` on L20, but I'm not sure what he means by "90% sure we are supposed to do gets for everything"? Maybe you could elaborate or clarify what you mean by that Raymond?

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -0,0 +1,50 @@
...
+        [HttpGet] // Create
+        public ActionResult<User> AddUser(User user)
+        {
+            if (user == null) return BadRequest();
```
joeriddles: Consider using `is null`

SecretSanta/src/SecretSanta.Api/DTO/Gift.cs
```diff
@@ -25,5 +18,25 @@ public Gift(SecretSanta.Domain.Models.Gift gift)
             OrderOfImportance = gift.OrderOfImportance;
             Url = gift.Url;
         }
+
+        public int Id { get; set; }
```
joeriddles: I'm curious as to why you moved the properties below the constructors?

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -29,19 +29,16 @@ public void GetGiftForUser_ReturnsUsersFromService()
...
-                }
-            };
+            var returnList = new List<Gift> {gift};
+
+            var testService = new TestableGiftService {GetGiftsForUser_Return = returnList};
+
             var controller = new GiftController(testService);
 
-            ActionResult<List<DTO.Gift>> result = controller.GetGiftForUser(4);
+            var result = controller.GetGiftForUser(4);
```
joeriddles: Why did you decide to use `var` throughout instead of the explicit types?

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -55,11 +52,143 @@ public void GetGiftForUser_RequiresPositiveUserId()
...
+            Assert.AreEqual(4, testService.AddGiftToUser_UserId);
+
+            // Ensure service was called
+            Assert.AreEqual(giftDto.Id, testService.AddGiftToUser_Gift.Id);
+        }
+
+        [TestMethod]
+        public void UpdateGiftForUser_RequiresGift()
+        {
+            var testService = new TestableGiftService();
```
joeriddles: Consider pulling duplicate code at the beginning of each test into a 
```
[TestInitialize]
public void TestInit()
{
...
}
```

SecretSanta/test/SecretSanta.Api.Tests/SecretSanta.Api.Tests.csproj
```diff
@@ -6,6 +6,14 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
joeriddles: Nice 👍

SecretSanta/test/SecretSanta.Api.Tests/TestableUserService.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public User UpdateUser(User user)
+        {
+            UpdateUser_User = user;
+
+            return UpdateUser_Return;
+        }
+
+        // Dont think this one is required for this assignment. Might come back to it regardless
```
joeriddles: Good use of a comment here.

SecretSanta/test/SecretSanta.Api.Tests/UserControllerTests.cs
joeriddles: I would actually somewhat disagree with @raymondshiner here. I think it makes more sense to explicitly specify when you're creating a `Model.User`/`Model.Gift`/etc. because this test lives in the `SecretSanta.Api.Tests` package, i.e. it lives in the same package name as the `DTO` objects, so I think it would make sense that the `DTO` objects are the ones that are used by default. Or be explicit everywhere you define an object that could be a `Model...` or `DTO...` 🤷‍♂️

SecretSanta/test/SecretSanta.Domain.Tests/Services/UserServiceTests.cs
```diff
@@ -21,8 +21,8 @@ public void AddUser_PersistsUser()
         {
             var user = new User
             {
-                FirstName = "John",
-                LastName = "Doe"
+                FirstName = "Kevin",
```
joeriddles: 👍

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
raymondshiner: @joeriddles I mean that although it wasn't in the Assignment 3 specifications. I am 90 % sure we are suppost to implement "Get" methods for all of our DTO Objects. in otherwords we should have a GetUser(int userId) and a GetGroup(int groupId) even though the Assignment3.md doesn't explicity say to.

SecretSanta/test/SecretSanta.Api.Tests/UserControllerTests.cs
raymondshiner: I can appreciate that, I agree that the important thing here though is be explicit about what you're using, because otherwise it just gets confusing and hard to read.

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
Keboo: At a minimum the assignment requires at least 3 "Get" methods. They are all listed under the required controller interaction with "Query *". Obviously if you want to implement individual "Get" methods for each of the DTOs that is certainly acceptable, but not explicitly required.

joeriddles: > we are suppost to implement "Get" methods for all of our DTO Objects. in otherwords we should have a GetUser(int userId) and a GetGroup(int groupId) even though the Assignment3.md doesn't explicity say to.

Aah that makes sense, thank you for clarifying what you meant. I was definitely misunderstanding.

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
joeriddles: As far as I understand, if the object being compared to `null` has an overloaded `Equals()` method, then using `== null` has the potential to not work correctly. Most of my reading on this question stems from this [post](https://stackoverflow.com/a/40676671). I was also under the impression it was in the coding guidelines to prefer `is null`, but after checking the guidelines, I do not see anything that supports my thinking. Maybe @Keboo could weigh in?

Keboo: @joeriddles that is a good post and I think it explains it quite well, but for those people who don't know CIL let me see if I can shed some light. Pre-C#7 (before the pattern matching feature that enables `is null` to work), you would check for `null` with either `== null` or `ReferenceEquals(user, null)` (if you [check the compiled IL](https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxASwDYB8ACAGAAhwEYBuAWACgqcBmIgJgIGECBvKgrg4Ae14wEAssQAUrAB4BKAgF4AfAQkE0UAgDsArhgwVKASD4DhDcUpkKlc2Ru27O3I4KG0z0uYoBKcAGaI46gDGcACiAI6aEBhQohJItjpSetwEDlxo6jBKydx0LKIZWdJsMAAWqgB0ErISpAC+aUT0JABsPPyCvAAOiBAwvAiyNma8xPGsvAwWiqNV1gSTVTlcea3txt29/QgAhMMTYywLU5azynvHS415vABuiAhoACZw64LhkdGivMAAVnCBLKTaYEUTiKTAuZDbLXeh3B7PV6FAgAcTgMAAEtBSsxeC9RCCJBU0Zjsbj8UkqHUgA=) you will notice this is the compiled equivalent of `is null`). The latter being if you wanted to make sure that poorly implemented overloaded operators did not break your code. 

There is only one case where this type of code would fail. In the hypothetical situation where someone have overridden the equality operator (`==`) and done something very very wrong.
Something like this for example:
```cs
public class User
{
    public static bool operator== (User user1, User user2) => false;
}
```
In this case the null check will return false (even if the user is null). In this [very very] rare case the `is null` works consistently while the `== null` fails.

So with all that said, should you favor `is null` (or `ReferenceEquals(user, null)`) instead of `== null`. 
[START PERSONAL OPINION] 
I would say the answer is no. IMO this falls under defensive programming. You will also notice the the built-in auto-code in both Visual Studio and R# simply insert `== null` when they generate the check. The rational is largely to trust that if someone does go to the effort of overloading the equality operator, they are going to do it in a sane manner. Now if you want to favor `is null` I am never going to say that is wrong because you are guaranteed to get more consistent behavior when checking for null.
[END PERSONAL OPINION]

SecretSanta/src/SecretSanta.Api/DTO/Gift.cs
Keboo: This is a good point. In general it is best to stay consistent with your file structure (ie. if you always put constructors above your properties do that for all classes). It is interesting that the Group class does not also follow this pattern.

SecretSanta/src/SecretSanta.Api/SecretSanta.Api.csproj
```diff
@@ -1,10 +1,18 @@
...
     <TargetFramework>netcoreapp2.2</TargetFramework>
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
Keboo: Pro tip: If you move `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` to the `<PropertyGroup>` (line 3) it will apply to all configurations and you wont need both of these (though I believe this is the way the Visual Studio GUI sets it if you do it from the there).

SecretSanta/src/SecretSanta.Domain/Models/Message.cs
```diff
@@ -6,11 +6,13 @@ public class Message
...
-        [ForeignKey("SenderId")]
-        public User Sender { get; set; }
+
+        [ForeignKey("SenderId")] public User Sender { get; set; }
+
         public int RecipientId { get; set; }
-        [ForeignKey("RecipientId")]
-        public User Recipient { get; set; }
+
+        [ForeignKey("RecipientId")] public User Recipient { get; set; }
```
Keboo: This is odd formatting. In general attributes are usually specified on their own line.

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -32,5 +33,62 @@ public List<Group> FetchAll()
...
+
+            DbContext.Groups.Remove(group);
+            DbContext.SaveChanges();
+        }
+
+        public User AddUserToGroup(int groupId, User user)
+        {
+            if (user == null) throw new ArgumentNullException(nameof(user));
+            
+            if (groupId <= 0) throw new ArgumentException($"Invalid groupId. {groupId} must be greater than 0");
```
Keboo: For `ArgumentException` (and its derived classes) you want to use the constructor overload that takes in the parameter name. 
```suggestion
            if (groupId <= 0) throw new ArgumentException($"Invalid {nameof(groupId)}. {groupId} must be greater than 0", nameof(groupId));
```

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -55,11 +52,152 @@ public void GetGiftForUser_RequiresPositiveUserId()
...
+                OrderOfImportance = 1,
+                Url = "Url"
+            };
+
+            var testService = new TestableGiftService();
+            var controller = new GiftController(testService);
+
+            var result = controller.UpdateGiftForUser(giftDto, -1);
+
+            Assert.IsTrue(result.Result is NotFoundObjectResult, $"{result.Result} is not instance of NotFoundResult");
```
Keboo:
```suggestion
            Assert.IsTrue(result.Result is NotFoundObjectResult, $"{result.Result} is not instance of NotFoundObjectResult");
```

SecretSanta/test/SecretSanta.Api.Tests/GroupControllerTests.cs
```diff
@@ -0,0 +1,318 @@
...
+            };
+
+            var testService = new TestableGroupService { AddGroup_Return = group };
+            var controller = new GroupController(testService);
+
+            var returnedAction = controller.CreateGroup(new DTO.Group(group)).Result;
+            
+            Assert.IsTrue(returnedAction is OkObjectResult);
+            
+            OkObjectResult result = returnedAction as OkObjectResult;
```
Keboo: This you have already verified the type, I would use a direct cast here rather than an `as` since that will redo the type check.

SecretSanta/test/SecretSanta.Api.Tests/GroupControllerTests.cs
```diff
@@ -0,0 +1,318 @@
...
+            ActionResult<DTO.User> result = controller.RemoveUserFromGroup(4, null);
+            
+            Assert.IsTrue(result.Result is BadRequestResult);
+            
+            // This check ensures that the service was not called
+            Assert.IsNull(testService.RemoveUserFromGroup_User);
+        }
+        
+        [TestMethod]
+        public void RemoveUserFromGroup_InvokesService() // TODO:
```
Keboo: Why is this comment here?

PR #66, opened 2019-01-30
Assignment 3 - Christopher Sturm

Sturminator:
Restarted after git issues, I understand how to properly branch now I think. I should be caught back up by the end of Wednesday. I don't recommend bothering with a CR until Wednesday evening.

### DIFF
--- a/SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
@@ -81,6 +81,9 @@ namespace SecretSanta.Api.Controllers
 
             var returnedUser = _GroupService.AddUserToGroup(groupId, DTO.User.ToDomainEntity(user));
 
+            if (returnedUser == null)
+                return NotFound();
+
             return new DTO.User(returnedUser);
         }
 
@@ -94,6 +97,9 @@ namespace SecretSanta.Api.Controllers
                 return BadRequest();
 
             var returnedUser = _GroupService.RemoveUserFromGroup(groupId, DTO.User.ToDomainEntity(user));
+
+            if (returnedUser == null)
+                return NotFound();
 
             return new DTO.User(returnedUser);
         }
--- a/SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
@@ -40,6 +40,17 @@ namespace SecretSanta.Api.Controllers
             return new DTO.Group(returnedGroup);
         }
 
+        [HttpDelete]
+        public ActionResult<DTO.Group> DeleteGroup(DTO.Group group)
+        {
+            if (group == null)
+                return BadRequest();
+
+            var returnedGroup = _GroupService.RemoveGroup(DTO.Group.ToDomainEntity(group));
+
+            return new DTO.Group(returnedGroup);
+        }
+
         [HttpGet]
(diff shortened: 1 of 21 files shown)


SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            if (groupId <= 0)
+                return NotFound();
+
+            if (user == null)
+                return BadRequest();
+
+            var returnedUser = _GroupService.RemoveUserFromGroup(groupId, user);
+
+            return new DTO.User(returnedUser);
```
BHoward85: Would you being checking if the returned user from the RemoveUserFromGroup for a null if the user is not in the group?
If not, this would be a case of passing a null.

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -0,0 +1,62 @@
...
+        [HttpPut]
+        public ActionResult<DTO.User> UpdateUser(DTO.User user)
+        {
+            if (user == null)
+                return BadRequest();
+
+            var returnedUser = _UserService.UpdateUser(DTO.User.ToDomainEntity(user));
+
+            return new DTO.User(returnedUser);
+        }
```
BHoward85: Same as above, would you UpdateUser have a check if the user is there and if not what would it pass back, this is something to think about.

SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
Sturminator: Nice catch, I'll make sure to throw in a check for that. Thanks!

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
Sturminator: Same as above, thanks!

Keboo: Though the implementation doesn't allow a null to be returned it is good to check for it. You never want to make assumptions about the implementation of an interface.

SecretSanta/test/SecretSanta.Api.Tests/TestableGiftService.cs
```diff
@@ -6,12 +6,41 @@ namespace SecretSanta.Api.Tests
 {
     public class TestableGiftService : IGiftService
     {
-        public List<Gift> ToReturn { get; set; }
+        public List<Gift> ToReturnList { get; set; }   
+        public Gift ToReturn { get; set; }
```
Keboo: This works for this assignment, but in general you want to have separate properties for the return values for each method. You can imagine unit testing a method that makes multiple calls against the same service and needing to control each call separately.

SecretSanta/test/SecretSanta.Api.Tests/UserControllerTests.cs
```diff
@@ -0,0 +1,156 @@
...
+            var testService = new TestableUserService
+            {
+                ToReturn = user
+            };
+
+            var controller = new UserController(testService);
+
+            ActionResult<DTO.User> result = controller.UpdateUser(new DTO.User(user));
+
+            Assert.IsNotNull(testService.UpdateUser_User);
```
Keboo: It would be good to check that the `testService.UpdateUser_User` at least matches the Id of the user you passed in. Just checking against null is a good start but not totally adequate.

PR #67, opened 2019-01-30
Assignment3

Isaac-Bliss:
had to scrap my previous work because I was doing it wrong, just started by trying to code to add a gift to the user

### DIFF
--- a/SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
@@ -18,7 +18,7 @@ namespace SecretSanta.Api.Controllers
             _GiftService = giftService ?? throw new ArgumentNullException(nameof(giftService));
         }
 
-        // GET api/Gift/5
+        // GET api/Gift/#
         [HttpGet("{userId}")]
         public ActionResult<List<DTO.Gift>> GetGiftForUser(int userId)
         {
@@ -29,6 +29,19 @@ namespace SecretSanta.Api.Controllers
             List<Gift> databaseUsers = _GiftService.GetGiftsForUser(userId);
 
             return databaseUsers.Select(x => new DTO.Gift(x)).ToList();
+            
+        }
+
+        // POST api/Gift/#
+        [HttpPost("{userId}")]
+        public ActionResult<DTO.Gift> CreateGift(int userId, DTO.Gift gift)
+        {
+            if (gift == null || userId <= 0)
+            {
+                return BadRequest();
+            }
+
+            return new DTO.Gift(_GiftService.AddGiftToUser(userId, DTO.Gift.ToDomain(gift)));
         }
     }
 }
--- a/SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
@@ -45,7 +45,7 @@ namespace SecretSanta.Api.Controllers
                 return BadRequest();
             }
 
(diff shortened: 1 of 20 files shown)


SecretSanta/src/SecretSanta.Api/Controllers/GiftController.cs
```diff
@@ -29,6 +29,19 @@ public ActionResult<List<DTO.Gift>> GetGiftForUser(int userId)
             List<Gift> databaseUsers = _GiftService.GetGiftsForUser(userId);
 
             return databaseUsers.Select(x => new DTO.Gift(x)).ToList();
+            
+        }
+
+        // POST api/Gift/#
+        [HttpPost("{userId}")]
+        public ActionResult<DTO.Gift> CreateGift(int userId, DTO.Gift gift)
```
Chester-Southwood: Not a issue with your code, but what I did in mine is name the objects based off of what class they originate from. Since there are two distinct Gift type objects, it may be a pain trying to see which one is which when testing. I named mine dtoGift to give a more "meaningful" name. Not really better since it's not that hard to see what type of gift it is, but I wanted to throw my two cents.

Isaac-Bliss:
Didn't implement the add/remove users from groups,
unit testing is very bad,
I still don't understand how the 'testable' services can give us useful information for testing.
Overall, this assignment was very hard to wrap my brain around, especially on how to go about testing, and creating test initialization.
I ran out of time, but hopefully I got some things correct.

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -27,6 +27,31 @@ public Group UpdateGroup(Group @group)
...
+            DbContext.SaveChanges();
+            return group;
+        }
+
+
+        public List<User> FetchGroupUsers(int groupId)
+        {
+            if (groupId <= 0)
+            {
+                throw new ArgumentException("groupId must be larger than 0 in FetchAllGroup");
```
Keboo: You typically do not need to include the method name in your exception message. All exceptions contain stack traces showing you exactly the chain of methods that led to the exception be thrown.

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -61,5 +61,60 @@ public void GetGiftForUser_RequiresPositiveUserId()
...
+        [TestMethod]
+        public void UpdateGiftForUser_NullGift()
+        {
+            var testService = new TestableGiftService();
+            var controller = new GiftController(testService);
+
+            ActionResult result = controller.UpdateGiftForUser(1, null);
+
+            Assert.IsTrue(result is BadRequestResult);
+        }
```
Keboo: It is great that you tested the failure cases. But part of the assignment was to use the testable service implementations to test the success case for each of the controller methods. That appears to be missing on all of the controllers.

SecretSanta/test/SecretSanta.Api.Tests/TestableGroupService.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        public Group AddGroup_Group { get; set; }
+        public Group ToReturnGroup { get; set; }
+        public Group AddGroup(Group group)
+        {
+            AddGroup_Group = group;
+            return ToReturnGroup;
+        }
+
+        public List<Group> ToReturnGroupList { get; set; }
```
Keboo: Though it works for this assignment, when writing these test classes you usually don't want to re-use the same property as the return value for multiple methods. You can imagine a controller making multiple calls to the same service and needing to test various combinations of return values from each of the methods.

PR #68, opened 2019-01-30
Assignment3 - Eric Fueng

efueng:
~~WARNING: This code is known to the State of California to cause macular degeneration. Wait to review until Wednesday night for your own safety.~~

Bug fix: code no longer causes macular degeneration


SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -1,65 +1,90 @@
...
+        }
+
+        [TestMethod]
+        public void AddGiftToUser_InvokesService()
+        {
+            var testService = new TestableGiftService();
+            var controller = new GiftController(testService);
+
+            ActionResult<DTO.Gift> result = controller.AddGiftToUser(4, new DTO.Gift());
+            OkResult okResult = result as OkResult;
```
COsborn2: This is what is causing your error (Im sure you know that). The reason why is because your `ActionResult<DTO.Gift>` contains a property called `Result` and that is the HTTP Code. So if instead of doing `OkResult okResult = result as OkResult` you can do `Assert.IsTrue(result.Result is OkResult)`, which I think is what you wanted to do. That way you're checking and making sure your service is executing properly.

SecretSanta/test/SecretSanta.Api.Tests/TestableGiftService.cs
```diff
@@ -1,18 +1,37 @@
...
+
+        public Gift AddGiftToUser(int userId, Gift gift)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public List<Gift> GetGiftsForUser(int userId)
+        {
+            GetGiftsForUser_UserId = userId;
+            GetGiftsForUser_Return = GetGiftsForUser(userId);
```
COsborn2: This is causing an infinite loop and a logger overflow if you try to run your tests. You can just do
```csharp
GetGiftsForUser_UserId = userId;
return GetGiftsForUser_Return;
```

If you need a little bit more explanation on what the TestableGiftService does or how to use it as validation of your Controller let me know.

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -1,65 +1,90 @@
...
+
+        [TestMethod]
+        public void GetGiftForUser_RequiresPositiveUserId()
+        {
+            var testService = new TestableGiftService();
+            var controller = new GiftController(testService);
+
+            ActionResult<List<DTO.Gift>> result = controller.GetGiftForUser(-1);
+            
+            Assert.IsTrue(result.Result is NotFoundResult);
```
COsborn2: You pass a String back with your `NotFound()` so this will actually be a `NotFoundObjectResult`

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -1,65 +1,90 @@
...
+
+        [TestMethod]
+        public void AddGiftToUser_RequiresGift()
+        {
+            var testService = new TestableGiftService();
+            var controller = new GiftController(testService);
+
+            ActionResult<DTO.Gift> result = controller.AddGiftToUser(4, null);
+
+            Assert.IsTrue(condition: result.GetType() == typeof(DTO.Gift));
```
COsborn2: In the same vein as the last test you're probably better off doing `Assert.IsTrue(result.Result is BadRequestObjectResult)`

SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
```diff
@@ -0,0 +1,69 @@
...
+    public class GroupController : ControllerBase
+    {
+        private readonly IGroupService _GroupService;
+
+        public GroupController(IGroupService groupService)
+        {
+            _GroupService = groupService ?? throw new ArgumentNullException(nameof(groupService));
+        }
+
+        private Group GroupDtoToEntity(DTO.Group dto)
```
COsborn2: You might want to move this to inside your DTO.Group

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class UserController : ControllerBase
+    {
+        private readonly IUserService _UserService;
+        
+        public UserController(IUserService userService)
+        {
+            _UserService = userService;
```
COsborn2: You should do a null check here like you do in your other controllers.

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
efueng: Thank you for this

SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
efueng: I thought about that and it's something I'm still debating. I figured since the controllers already know about both the Model and DTO objects it might be better in here. My reasoning is that otherwise DTOs would have to know about their Model counterparts. Thoughts?

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
efueng: Done

SecretSanta/test/SecretSanta.Api.Tests/TestableGiftService.cs
efueng: Thanks Cameron. The testable interfaces were a bit weird for me. I ended up doing mock objects which was... a bit less confusing lol.

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
efueng: Actually I did not know that. Good catch and thank you.

SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
Keboo: IMO either is fine. My personal preference is extension methods for this type of stuff, but that is just me.

SecretSanta/src/SecretSanta.Api/DTO/Gift.cs
```diff
@@ -1,29 +1,34 @@
...
+            if (gift == null) throw new ArgumentNullException(nameof(gift));
+
+            Id = gift.Id;
+            Title = gift.Title;
+            Description = gift.Description;
+            OrderOfImportance = gift.OrderOfImportance;
+            Url = gift.Url;
+        }
+
+        public static Gift ToDTO(int userId, SecretSanta.Domain.Models.Gift gift)
```
Keboo: Method appears un-used

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -1,36 +1,117 @@
...
+
+            DbContext.Remove(@group);
+            DbContext.SaveChanges();
+        }
+
+        public User AddGroupUser(int groupId, User user)
+        {
+            if (groupId <= 0)
+            {
+                throw new ArgumentException("groupId <= 0 on call to GroupService.AddGroupUser(int groupId, User user).",
```
Keboo: Exceptions will always contain a stack trace showing the chain of methods that led to the exception being thrown. Because of this, we don't include the method name/signature as part of the message.

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -1,65 +1,223 @@
...
+        {
+            new GiftController(null);
+        }
+
+        [TestMethod]
+        public void GetGiftForUser_ReturnsGiftsFromService()
+        {
+            var mocker = new AutoMocker(MockBehavior.Strict);
+            var service = mocker.GetMock<IGiftService>();
+            var controller = new GiftController(service.Object);
```
Keboo: Part of the advantage of using AutoMocker is not needing to invoke the controller's constructor directly (which makes you manually specify all of the constructor parameters). If you let AtuoMocker create the instances it will populate the constructor parameters for you, making your test more resilient to constructor signature changes.
```suggestion
            var controller = mocker.CreateInatance<GiftController>();
```

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -1,65 +1,223 @@
...
+            var giftMock = mocker.CreateInstance<Gift>();
```
Keboo: Typically for things like models and DTOs you would not use AutoMapper to create them. This is little more than a light wrapper over invoking the constructor directly.

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -1,65 +1,223 @@
...
+            service.Setup(x => x.GetGiftsForUser(giftMock.UserId))
+                .Returns(new List<Gift>())
+                .Verifiable();
+
+            // Act
+            var result = controller.GetGiftForUser(giftMock.UserId);
+
+            // Assert
+            Assert.IsTrue(result.Value is List<DTO.Gift>);
+            service.Verify(x => x.GetGiftsForUser(giftMock.UserId), Times.Once);
```
Keboo: I would actually advocate dropping this Verify call in favor of beefing up the Setup call instead. 
The above setup will only match if the user id matches. But I would make it actually return a gift in the list and then check to make sure the returned value from the controller contains a gift that matches the values.

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -1,65 +1,223 @@
...
+            // Assert
+            Assert.IsTrue(result.Result is NotFoundResult);
+            service.Verify(x => x.GetGiftsForUser(-1), Times.Never);
+        }
+
+        [TestMethod]
+        public void GetGiftsForUser_ReturnsListOfGifts()
+        {
+            // Arrange
+            var gift = Mock.Of<Gift>();
```
Keboo: I would not use a Mock here, for simple POCOs just invoke the constructor directly.

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -1,65 +1,223 @@
...
+            Assert.AreEqual(gift.Url, resultDTO.Url);
+            Assert.AreEqual(gift.OrderOfImportance, resultDTO.OrderOfImportance);
+            Mocker.VerifyAll();
+            GiftServiceMock.VerifyAll();
+        }
+
+        [TestMethod]
+        public void AddGiftToUser_RequiresGift()
+        {
+            var controllerMock = Mocker.CreateInstance<GiftController>();
```
Keboo: :+1:

SecretSanta/test/SecretSanta.Api.Tests/GiftControllerTests.cs
```diff
@@ -1,65 +1,223 @@
...
+            var result = controllerMock.AddGiftToUser(1, null);
+            Assert.IsTrue(result is BadRequestResult);
+            
+        }
+
+        [TestMethod]
+        public void AddGiftToUser_NegativeUserId_ReturnsNotFound()
+        {
+            // Arrange
+            var strictMocker = new AutoMocker(MockBehavior.Strict);
```
Keboo: Nice use of strict mocks

SecretSanta/test/SecretSanta.Api.Tests/GroupControllerTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            service.Setup(x => x.RemoveGroup(It.IsAny<Domain.Models.Group>())).Verifiable();
+
+            // Act
+            var result = controller.RemoveGroup(null);
+
+            // Assert
+            Assert.IsTrue(result is BadRequestResult);
+            service.Verify(x => x.UpdateGroup(null), Times.Never);
+        }
+
```
Keboo: I think you forgot to test to success case for RemoveGroup

PR #72, opened 2019-02-07
Assignment4 - Raymond Shiner

raymondshiner:


SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
```diff
@@ -14,52 +16,59 @@ namespace SecretSanta.Api.Controllers
     public class GroupController : ControllerBase
```
breaman: This class should have been decorated with [ApiController] to light up the analysis issues.

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -2,8 +2,10 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using AutoMapper;
 using Microsoft.AspNetCore.Mvc;
 using SecretSanta.Api.ViewModels;
+using SecretSanta.Domain.Models;
 using SecretSanta.Domain.Services.Interfaces;
 
 // For more information on enabling Web API for empty projects, visit https://go.microsoft.com/fwlink/?LinkID=397860
```
breaman: Same as GroupController, this needed to be decorated with the [ApiController] attribute. Since this wasn't there, you didn't see any warnings, and this is why GiftController had so many more attribute decorations on the methods.

SecretSanta/src/SecretSanta.Domain/Models/ApplicationDbContext.cs
```diff
@@ -11,7 +11,7 @@ public class ApplicationDbContext : DbContext
         public DbSet<Pairing> Pairings { get; set; }
         public ApplicationDbContext(DbContextOptions<ApplicationDbContext> options) : base(options)
         {
-            Database.EnsureCreated();
+            
         }
 
         protected override void OnModelCreating(ModelBuilder modelBuilder)
```
breaman: Where is the call to make Name unique for the Group object? The only way to do that is to use the fluent api.
```
modelBuilder.Entity<Group>().HasIndex(g => g.Name).IsUnique();
```

PR #73, opened 2019-02-07
Assignment4

grantdwoods:


grantdwoods:
My Post requests return an empty string url (location). I did not understand the context of "nameof(Get)" at the time and I meant to ask about it/fix it before submitting. Clearly I forgot about it. Just leaving a comment here to save the review comments on all the post requests.

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -2,41 +2,45 @@
...
-            if (userViewModel == null)
+            if (userInputViewModel == null)
             {
                 return BadRequest();
             }
 
-            var persistedUser = UserService.AddUser(UserInputViewModel.ToModel(userViewModel));
-
-            return Ok(UserViewModel.ToViewModel(persistedUser));
+            return Created("",
```
breaman: Would have been better to use the CreatedAtAction method here so that you could have passed back the proper url for "fetching" the user back out.

SecretSanta/src/SecretSanta.Api/Models/GroupMapperProfile.cs
```diff
@@ -0,0 +1,21 @@
...
+using SecretSanta.Domain.Models;
+
+namespace SecretSanta.Api.Models
+{
+    public class GroupMapperProfile : Profile
+    {
+        public GroupMapperProfile()
+        {
+            CreateMap<Group, GroupViewModel>();
+            CreateMap<Group, GroupInputViewModel>();
```
breaman: Nothing wrong per se with creating this one, but it is a Map that is never used and never should be used.

SecretSanta/src/SecretSanta.Api/Models/GroupMapperProfile.cs
```diff
@@ -0,0 +1,21 @@
...
+            CreateMap<GroupInputViewModel, Group>();
+            CreateMap<GroupViewModel, GroupInputViewModel>();
```
breaman: Is this map used? I would imagine this could have been eliminated as well.

SecretSanta/src/SecretSanta.Api/Models/UserMapperProfile.cs
```diff
@@ -0,0 +1,25 @@
...
+        public UserMapperProfile()
+        {
+            CreateMap<User, UserViewModel>();
+
+            CreateMap<User, UserInputViewModel>()
+                .ForSourceMember(x => x.Id, opt => opt.DoNotValidate());
+
+            CreateMap<UserInputViewModel, User>();
+
+            CreateMap<UserViewModel, UserInputViewModel>();
```
breaman: Again, this map is probably never used (nor should be). It is best practice to not create maps till you actually need them. Not sure if there is a performance issue around it, though.

SecretSanta/src/SecretSanta.Api/Models/UserMapperProfile.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace SecretSanta.Api.Models
+{
+    public class UserMapperProfile : Profile
+    {
+        public UserMapperProfile()
+        {
+            CreateMap<User, UserViewModel>();
+
+            CreateMap<User, UserInputViewModel>()
+                .ForSourceMember(x => x.Id, opt => opt.DoNotValidate());
```
breaman: Another unnecessary map.

PR #74, opened 2019-02-07
Assignment4

Chester-Southwood:
Initial push, most of functionality implemented. Further testing needed.


SecretSanta/src/SecretSanta.Api/Models/AutoMapperProfileConfiguration.cs
```diff
@@ -0,0 +1,36 @@
...
+        public AutoMapperProfileConfiguration()
+        {
+            CreateMap<UserInputViewModel, User>();
+            CreateMap<User, UserInputViewModel>();
+
+            CreateMap<User, UserViewModel>();
+            CreateMap<UserViewModel, User>();
+
+            CreateMap<GroupInputViewModel, Group>();
+            CreateMap<Group, GroupInputViewModel>();
```
breaman: It is usually best practice to only create the maps you need/use. Group -> GroupInputViewModel and User -> UserInputViewModel is never used. Also, the UserViewModel -> User and GroupViewModel -> Group are also not used in our situations.

SecretSanta/src/SecretSanta.Api/Startup.cs
```diff
@@ -5,41 +5,53 @@
...
+using Microsoft.Extensions.DependencyModel;
 using SecretSanta.Domain.Models;
 using SecretSanta.Domain.Services;
 using SecretSanta.Domain.Services.Interfaces;
 using Swashbuckle.AspNetCore.Swagger;
+using System.Linq;
+using System.Reflection;
+using AutoMapper;
 
+[assembly: ApiConventionType(typeof(DefaultApiConventions))]
```
breaman: You didn't need to add both the ProducesResponseType and this global attribute. If you wanted to use this, you should have gone in and removed the unnecessary ProducesResponseType attributes on your controllers.

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -1,11 +1,14 @@
+using System.ComponentModel.DataAnnotations;
+
 namespace SecretSanta.Domain.Models
 {
     public class Gift
     {
         public int Id { get; set; }
+        [Required(AllowEmptyStrings = false)]
```
breaman: You could have left off the AllowEmptyStrings = false, that is the default value for that parameter.

SecretSanta/src/SecretSanta.Domain/SecretSanta.Domain.csproj
```diff
@@ -5,6 +5,9 @@
   </PropertyGroup>
 
   <ItemGroup>
+      <PackageReference Include="AutoMapper" Version="8.0.0" />
+      <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="6.0.0" />
+      <PackageReference Include="Microsoft.AspNetCore.Mvc.Api.Analyzers" Version="2.2.0" />
```
breaman: The analyzer and AutoMapper references did not have to be put into this project. They are only used in the Api project.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/GitfControllerTests.cs
```diff
@@ -48,13 +58,45 @@ public void GetGiftForUser_ReturnsUsersFromService()
...
+            IActionResult result = controller.GetGiftForUser(-1);
 
-            Assert.IsTrue(result.Result is NotFoundResult);
+            Assert.IsNotNull(result);
             //This check ensures that the service was not called
             Assert.AreEqual(0, testService.GetGiftsForUser_UserId);
         }
+
+        [TestMethod]
+        public void CreateGift_CompletesUnsuccessfully_EmptyName()
```
breaman: Interesting extra set of tests to create here. Since the validation for the Title was done at the database level, this would have been a better test to have against the service that adds a gift to the database. We don't have that service at the moment, so this is a clever way to do this as well.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/GroupControllerTests.cs
```diff
@@ -219,13 +213,44 @@ public void AddUserToGroup_ReturnsOkWhenUserAddedToGroup()
...
-            var controller = new GroupController(service.Object);
+            var controller = new GroupController(service.Object, Mapper.Instance);
 
-            ActionResult result = controller.AddUserToGroup(2, 3);
+            Assert.IsNotNull(controller.AddUserToGroup(2, 3) as OkObjectResult);
+        }
 
-            Assert.IsTrue(result is OkResult);
+        [TestMethod]
+        public void CreateGroup_CompletesUnsuccessfully_EmptyName()
```
breaman: This one should have been similar to your CreateGroup test that called `controller.CreateGroup(group) as CreatedAtActionResult;` and verified the group didn't actually get created. It is nice to see you test to make sure it passes validation, but we want to verify it can't be persisted in the database either, not just fail code validation.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/UserControllerTests.cs
```diff
@@ -1,10 +1,46 @@
...
+using SecretSanta.Api.ViewModels;
+using System;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Text;
 
 namespace SecretSanta.Api.Tests.Controllers
 {
-    class UserControllerTests
+    [TestClass]
```
breaman: You should have had some tests in here that used the `CustomWebApplicationFactory` class that we created in during class. The tests then should have created a "fake client" and called the api through the actual route, not just via the method. Similar to `await client.PostAsJsonAsync("/api/users", viewModel);`

PR #75, opened 2019-02-07
Assignment4

tdang2018:
Making a Pull Request


SecretSanta/src/SecretSanta.Api/Models/AutoMapperProfileConfiguration.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Threading.Tasks;
+
+namespace SecretSanta.Api.Models
+{
+    public class AutoMapperProfileConfiguration : Profile
+    {
+        public AutoMapperProfileConfiguration()
+        {
+            CreateMap<UserInputViewModel, User>();
+            CreateMap<User, UserInputViewModel>();
```
breaman: This map is not needed, the code doesn't use anything where the User object needs to be mapped to UserInputViewModel.

SecretSanta/src/SecretSanta.Api/Models/AutoMapperProfileConfiguration.cs
```diff
@@ -0,0 +1,28 @@
...
+
+            CreateMap<User, UserViewModel>();
+
+            CreateMap<GroupInputViewModel, Group>();
+            CreateMap<Group, GroupInputViewModel>();
```
breaman: This map is also not necessary.

PR #76, opened 2019-02-08
Assignment 4

Isaac-Bliss:
first push

### DIFF
--- a/SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
@@ -2,9 +2,11 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using AutoMapper;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using SecretSanta.Api.ViewModels;
+using SecretSanta.Domain.Models;
 using SecretSanta.Domain.Services.Interfaces;
 
 // For more information on enabling Web API for empty projects, visit https://go.microsoft.com/fwlink/?LinkID=397860
@@ -15,10 +17,12 @@ namespace SecretSanta.Api.Controllers
     public class GroupController : ControllerBase
     {
         private IGroupService GroupService { get; }
+        private IMapper Mapper { get; }
 
-        public GroupController(IGroupService groupService)
+        public GroupController(IGroupService groupService, IMapper mapper)
         {
             GroupService = groupService ?? throw new ArgumentNullException(nameof(groupService));
+            Mapper = mapper;
         }
 
         // GET api/group
@@ -27,7 +31,7 @@ namespace SecretSanta.Api.Controllers
         [ProducesDefaultResponseType]
         public ActionResult<IEnumerable<GroupViewModel>> GetAllGroups()
         {
-            return new ActionResult<IEnumerable<GroupViewModel>>(GroupService.FetchAll().Select(x => GroupViewModel.ToViewModel(x)));
+            return new ActionResult<IEnumerable<GroupViewModel>>(GroupService.FetchAll().Select(x => Mapper.Map<GroupViewModel>(x)));
         }
 
         // POST api/group
@@ -42,7 +46,7 @@ namespace SecretSanta.Api.Controllers
                 return BadRequest();
             }
 
-            return GroupViewModel.ToViewModel(GroupService.AddGroup(GroupInputViewModel.ToModel(viewModel)));
+            return Mapper.Map<GroupViewModel>(Mapper.Map<Group>(viewModel));
         }
 
         // PUT api/group/5
@@ -65,7 +69,7 @@ namespace SecretSanta.Api.Controllers
 
             fetchedGroup.Name = viewModel.Name;
 
-            return GroupViewModel.ToViewModel(GroupService.UpdateGroup(fetchedGroup));
+            return Mapper.Map<GroupViewModel>(GroupService.UpdateGroup(fetchedGroup));
         }
 
         [HttpPut("{groupId}/{userid}")]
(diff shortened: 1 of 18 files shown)


Isaac-Bliss:
AutoMapper has a least a couple problems I don't know how/where to fix, so that's causing my unit tests to fail. I was unable to create a migration, as I received the error "There is no runtime associated with this framework, and projects targeting it cannot be executed directly". I think I made the Sqlite connection to use a physical connection string, but I probably missed a piece somewhere. 
Other than that, I tried to follow the assignment as best as possible, A lot of stuff went over my head, but I made my best attempt on everything. 
I am tired, and spinning my wheels is getting nowhere, I'll take my lumps and try to do better on the next assignment, thanks for reading

SecretSanta/src/SecretSanta.Api/SecretSanta.Api.csproj
```diff
@@ -6,6 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="6.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.App" />
     <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="2.2.0" PrivateAssets="All" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="4.0.1" />
```
breaman: You are missing the analyzer package, so there is no way to validate through the compiler if your API is documented correctly.

SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
```diff
@@ -14,34 +17,45 @@ namespace SecretSanta.Api.Controllers
     public class GroupController : ControllerBase
```
breaman: This class needed to be decorated with the [ApiController] attribute in order to get proper default behavior.

SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
```diff
@@ -14,34 +17,45 @@ namespace SecretSanta.Api.Controllers
...
+        [ProducesDefaultResponseType]
+        public IActionResult CreateGroup(GroupInputViewModel viewModel)
         {
             if (viewModel == null)
             {
                 return BadRequest();
             }
 
-            return GroupViewModel.ToViewModel(GroupService.AddGroup(GroupInputViewModel.ToModel(viewModel)));
+            return Ok(Mapper.Map<GroupViewModel>(Mapper.Map<Group>(viewModel)) );
```
breaman: Post is suppose to return Created, not OK

SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
```diff
@@ -55,11 +69,15 @@ public ActionResult<GroupViewModel> UpdateGroup(int id, GroupInputViewModel view
 
             fetchedGroup.Name = viewModel.Name;
 
-            return GroupViewModel.ToViewModel(GroupService.UpdateGroup(fetchedGroup));
+            return Ok(Mapper.Map<GroupViewModel>(GroupService.UpdateGroup(fetchedGroup)) );
```
breaman: Put is supposed to return NoContent instead of Ok

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -14,29 +17,38 @@ namespace SecretSanta.Api.Controllers
     public class UserController : ControllerBase
```
breaman: This class also needed to be decorated with the [ApiController] attribute.

SecretSanta/src/SecretSanta.Domain/Models/ApplicationDbContext.cs
```diff
@@ -10,9 +10,7 @@ public class ApplicationDbContext : DbContext
         public DbSet<Message> Messages { get; set; }
         public DbSet<Pairing> Pairings { get; set; }
         public ApplicationDbContext(DbContextOptions<ApplicationDbContext> options) : base(options)
-        {
-            Database.EnsureCreated();
-        }
+        {}
 
         protected override void OnModelCreating(ModelBuilder modelBuilder)
         {
```
breaman: The method needed to be updated here so that you could make the name of the group unique in the database.

SecretSanta/src/SecretSanta.Domain/Models/Gift.cs
```diff
@@ -1,10 +1,14 @@
...
 namespace SecretSanta.Domain.Models
 {
     public class Gift
     {
         public int Id { get; set; }
+        [Required]
         public string Title { get; set; }
         public string Description { get; set; }
+
         public int OrderOfImportance { get; set; }
```
breaman: the assignment wanted OrderOfImportance to not be required, which means this needed to be `public int?` instead of `public int`.

SecretSanta/src/SecretSanta.Domain/Models/Group.cs
```diff
@@ -1,10 +1,13 @@
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 
 namespace SecretSanta.Domain.Models
 {
     public class Group
     {
         public int Id { get; set; }
+        [Required]
+        [Key]
```
breaman: The Id should be the primary key, so you don't want the [Key] attribute here on Name.

SecretSanta/src/SecretSanta.Domain/SecretSanta.Domain.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+      <PackageReference Include="Microsoft.AspNetCore.Mvc.Api.Analyzers" Version="2.2.0" />
```
breaman: This needed to be added to the Api project, not the domain project.

breaman:
Take a look at https://github.com/IntelliTect-Samples/EWU-CSCD496-2019-Winter/tree/Assignment4_StokesSolution to see what needed to be done for AutoMapper and unit testing (in particular, look at the GiftControllerTest and the method decorated with the AssemblyInitialize attribute.

PR #77, opened 2019-02-08
Assignment4

RichardTeller:


SecretSanta/src/SecretSanta.Api/Models/AutoMapperProfileConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading.Tasks;
+
+namespace SecretSanta.Api.Models
+{
+    public class AutoMapperProfileConfiguration : Profile
+    {
+        public AutoMapperProfileConfiguration()
+        {
+            CreateMap<UserInputViewModel, User>();
+            CreateMap<User, UserInputViewModel>();
```
breaman: This map is not needed.

breaman:
Missed a few things around documenting the api's correctly for Swagger. Also made the same mistake I made in class about pulling down the Mvc.Analyzers instead of the Mvc.Api.Analyzers package

PR #78, opened 2019-02-08
Assignment 4 - Christopher Sturm

Sturminator:


breaman:
Overall nice work. Missed a couple of requirements, but the work done was good.

PR #85, opened 2019-02-08
Assignment4 - Eric Fueng

efueng:
## Paired with @COsborn2 

**Shortcomings:**
* Couldn't get null first name to return BadRequest in UserControllerTests
* Assembly level ApiConventionType decoration doesn't work for all controller methods


SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -1,69 +1,85 @@
...
+using AutoMapper;
+using Microsoft.AspNetCore.Mvc;
+using SecretSanta.Api.ViewModels;
+using SecretSanta.Domain.Models;
+using SecretSanta.Domain.Services.Interfaces;
+
+// For more information on enabling Web API for empty projects, visit https://go.microsoft.com/fwlink/?LinkID=397860
+namespace SecretSanta.Api.Controllers
+{
+    [Route("api/[controller]")]
```
breaman: Missing the [ApiController] attribute on this. Having this attribute is what lights up the api analyzer. This also would have fixed the unit testing issues you ran into for requiring FirstName

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -1,69 +1,85 @@
...
+        {
+            if (userViewModel == null)
+            {
+                return BadRequest();
+            }
+
+            var persistedUser = UserService.AddUser(Mapper.Map<User>(userViewModel));
+            //var persistedUser = UserService.AddUser(UserInputViewModel.ToModel(userViewModel));
+
+            return Created("", Mapper.Map<UserViewModel>(persistedUser));
```
breaman: Should have used the CreatedAtAction method here so you could have auto generated the url instead of returning an empty string here.

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
```diff
@@ -1,69 +1,85 @@
...
+                return NotFound();
+            }
+
+            foundUser.FirstName = userViewModel.FirstName;
+            foundUser.LastName = userViewModel.LastName;
+
+            var persistedUser = UserService.UpdateUser(foundUser);
+            //var persistedUser = UserService.UpdateUser(foundUser);
+
+            return Ok(Mapper.Map<UserViewModel>(persistedUser));
```
breaman: Put is supposed to return NoContent instead of Ok

SecretSanta/src/SecretSanta.Api/Controllers/GroupController.cs
```diff
@@ -1,97 +1,118 @@
...
+            }
+            var fetchedGroup = GroupService.Find(id);
+            if (fetchedGroup == null)
+            {
+                return NotFound();
+            }
+
+            fetchedGroup.Name = viewModel.Name;
+
+            return Created(nameof(PutUpdateGroup), Mapper.Map<GroupViewModel>(GroupService.UpdateGroup(fetchedGroup)));
```
breaman: This should return a NoContent instead of Created.

SecretSanta/src/SecretSanta.Api/Models/AutoMapperProfileConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace SecretSanta.Api.Models
+{
+    public class AutoMapperProfileConfiguration : Profile
+    {
+        public AutoMapperProfileConfiguration()
+        {
+            CreateMap<Gift, GiftViewModel>();
+            
+            // Ignores User.Id in lossy conversion to UserInputViewModel
+            CreateMap<User, UserInputViewModel>()
```
breaman: This map is actually not needed.

SecretSanta/src/SecretSanta.Domain/Migrations/20190208071748_UpdatedUserViewModelConstraints.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace SecretSanta.Domain.Migrations
+{
+    public partial class UpdatedUserViewModelConstraints : Migration
+    {
```
breaman: This migration is empty, it should have had the extra requirements in it (but it looks like those were all made in the initial migration).

SecretSanta/src/SecretSanta.Api/Controllers/UserController.cs
efueng: Oh wow that was driving me crazy. Thanks for catching that!

SecretSanta/src/SecretSanta.Api/Models/AutoMapperProfileConfiguration.cs
efueng: Why is that? Is it because you don't lose information when mapping UserInputViewModel to User? So can we just cast UserInputViewModel to User?

PR #86, opened 2019-02-08
Assignment4

wpcwilliams:
May update testing if I have time (with late tag)


breaman:
Overall not bad. Looks like you were missing a couple of attributes and just had some issues with naming things which caused you some issues. Also looks like you didn't run a complete migration after changing your Model classes (don't see where Gift.Name, Gift.OrderOfImportance and Group.Name modified in the database). You got the models annotated correctly, but never got the changes into a migration.

PR #89, opened 2019-02-11
Assignment5

KyleBurgi:


SecretSanta/test/SecretSanta.Api.Tests/Controllers/GroupControllerTests.cs
```diff
@@ -49,36 +51,36 @@ public void GetAllGroups_ReturnsGroups()
...
+        public async Task CreateGroup_ReturnsCreatedGroup()
         {
             var group = new GroupInputViewModel
             {
                 Name = "Group"
             };
             var service = new Mock<IGroupService>();
             service.Setup(x => x.AddGroup(It.Is<Group>(g => g.Name == group.Name)))
-                .Returns(new Group
+                .Returns(Task.FromResult(new Group
```
Keboo: An alternative here would be to leverage the `.ReturnsAsync` extension method.
```suggestion
                .ReturnsAsync(new Group
```

SecretSanta/test/SecretSanta.Api.Tests/Controllers/GiftControllerTests.cs
```diff
@@ -1,68 +1,69 @@
...
+            {
+                Id = 3,
+                Title = "Gift Tile",
+                Description = "Gift Description",
+                Url = "http://www.gift.url",
+                OrderOfImportance = 1
+            };
+            TestableGiftService testService = new TestableGiftService
+            {
+                ToReturn = new List<Gift>
```
Keboo: There is actually a compile issue here since the `ToReturn` property was updated to be `Task<List<Gift>>`

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            await DbContext.Pairings.AddRangeAsync(myPairings);
+            await DbContext.SaveChangesAsync();
+
+
+            return true;
+        }
+
+        private List<Pairing> GetPairings(List<int> userIds)
+        {
```
Keboo: This is a start, but the assignment called for the pairings to be randomized.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,66 @@
...
+using SecretSanta.Domain.Models;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SecretSanta.Domain.Services
+{
+    public class PairingService : IPairingService
```
Keboo: This was a good start, but the assignment called for creating a `PairingController` that used this service.

Keboo: It would have been nice to see some unit test for this class.

PR #90, opened 2019-02-12
Assignment5 - Eric Fueng

efueng:


SecretSanta/src/SecretSanta.Api/Controllers/GroupsController.cs
```diff
@@ -49,46 +51,46 @@ public IActionResult Get(int id)
...
         [HttpPut]
-        public IActionResult Put(int id, GroupInputViewModel viewModel)
+        public async Task<IActionResult> Put(int id, GroupInputViewModel viewModel)
         {
             if (viewModel == null)
             {
                 return BadRequest();
             }
-            var group = GroupService.GetById(id);
+            Group group = await GroupService.GetById(id);
```
COsborn2: It might just be me defaulting back to the pre-C# dark ages, but I definitely prefer what you did here. I like explicitly stating my variable type instead of just using var, like you did here.

SecretSanta/src/SecretSanta.Api/Controllers/PairingsController.cs
```diff
@@ -0,0 +1,44 @@
...
+        public async Task<IActionResult> Get(int groupId)
+        {
+            if (groupId <= 0)
+            {
+                return BadRequest();
+            }
+
+            if (await PairingService.GeneratePairings(groupId))
+            {
+                return Ok();
```
COsborn2: You probably know this, but you're not actually doing anything with your Mapper in this class. You could do it like in the other controllers and map your view model with the Mapper and return that inside of you Ok().

SecretSanta/src/SecretSanta.Api/Controllers/UsersController.cs
```diff
@@ -28,16 +28,22 @@ public UsersController(IUserService userService, IMapper mapper)
         // GET api/User
         [HttpGet]
         [Produces(typeof(ICollection<UserViewModel>))]
-        public IActionResult Get()
+        public async Task<IActionResult> Get()
         {
-            return Ok(UserService.FetchAll().Select(x => Mapper.Map<UserViewModel>(x)));
+            List<User> users = await UserService.FetchAll();
+            return Ok(users.Select(x => Mapper.Map<UserViewModel>(x)));
+            //return Ok(UserService.FetchAll().Select(x => Mapper.Map<UserViewModel>(x)));
```
COsborn2: Im sure these comments are just leftover from converting your methods into async/await, but you should probably remove this before the final PR.

SecretSanta/src/SecretSanta.Api/SecretSanta.Api.csproj
```diff
@@ -11,6 +11,7 @@
     <PackageReference Include="Microsoft.AspNetCore.App" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Api.Analyzers" Version="2.2.0" />
     <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="2.2.0" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.2.0" />
```
COsborn2: You should turn on warnings as errors in these csproj files.

SecretSanta/src/SecretSanta.Domain/Services/Interfaces/IGroupUserService.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+// For more information on enabling Web API for empty projects, visit https://go.microsoft.com/fwlink/?LinkID=397860
+
+namespace SecretSanta.Api.Services.Interfaces
+{
+    public interface IGroupUserService
+    {
+        //TODO
```
COsborn2: I haven't quite gotten here yet, but I dont think you need to do this for this assignment. Definitely could be wrong though.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,77 @@
...
+	        */
+            Task<List<Pairing>> task = Task.Run(() => GetPairings(userIds));
+            List<Pairing> pairings = await task;
+
+            await DbContext.AddRangeAsync();
+            await DbContext.SaveChangesAsync();
+
+            return true;
+        }
+
```
COsborn2: Make sure when you implement Random on this you abstract away the Random calls so you can test it.

SecretSanta/test/SecretSanta.Api.Tests/TestableGiftService.cs
```diff
@@ -11,10 +12,10 @@ public class TestableGiftService : IGiftService
         public List<Gift> ToReturn { get; set; }
         public int GetGiftsForUser_UserId { get; set; }
 
-        public List<Gift> GetGiftsForUser(int userId)
+        public Task<List<Gift>> GetGiftsForUser(int userId)
         {
             GetGiftsForUser_UserId = userId;
-            return ToReturn;
+            return Task.FromResult(ToReturn);
```
COsborn2: I like this better than what I did. If you make the method async it will autobox your return type into a Task, but then you get a compiler warning about not using await inside your async method. Your way is definitely better.

SecretSanta/src/SecretSanta.Api/SecretSanta.Api.csproj
```diff
@@ -3,6 +3,12 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp2.2</TargetFramework>
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
+    <UserSecretsId>b31616e5-ba5b-4575-b6e2-2a5731362f36</UserSecretsId>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
Keboo: This is good (and the default way Visual Studio sets this property if you do it in the GUI), but this only turns on Warnings as Errors when you are in Debug mode. If you move `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` into the PropertyGroup above (the one without a Condition), it will apply to all configurations.

SecretSanta/src/SecretSanta.Domain/Models/ThreadSafeRandom.cs
```diff
@@ -0,0 +1,38 @@
...
+ * Heavily influenced by https://stackoverflow.com/questions/3049467/is-c-sharp-random-number-generator-thread-safe
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Models
+{
+    public class ThreadSafeRandom
```
Keboo: Very clever implementation! Given that the two fields being used are static, you could also make this entire class static.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SecretSanta.Domain.Services
+{
+    public class PairingService : IPairingService
+    {
+        private ApplicationDbContext DbContext { get; set; }
+        private ThreadSafeRandom Random { get; set; }
```
Keboo: These properties could be readonly.
```suggestion
        private ThreadSafeRandom Random { get; }
```

SecretSanta/test/SecretSanta.Api.Tests/Controllers/GiftControllerTests.cs
```diff
@@ -53,16 +56,72 @@ public void GetGiftForUser_ReturnsUsersFromService()
...
 
             Assert.IsNotNull(result);
             //This check ensures that the service was not called
             Assert.AreEqual(0, testService.GetGiftsForUser_UserId);
         }
+
+        [TestMethod]
+        [DataRow(-1, null)]
+        [DataRow(0, null)]
+        [DataRow(1, null)]
```
Keboo: :+1:

SecretSanta/test/SecretSanta.Api.Tests/TestableGiftService.cs
Keboo: As long as you are using "autobox" to mean "wrap inside of another object" and not [C# boxing](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/types/boxing-and-unboxing) then yes I agree. In fact if you make the method async (without the await) just so you can return a value then the compiler will simply wrap the result into a Task anyway, so you haven't really saved anything.

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,148 @@
...
+            {
+                PairingService pairingService = new PairingService(context);
+                List<Pairing> pairings1 = await pairingService.GeneratePairings(1);
+                List<Pairing> pairings2 = await pairingService.GeneratePairings(2);
+
+                List<int> pairing1UserIds = pairings1.Select(x => x.SantaId).ToList();
+                List<int> pairing2UserIds = pairings2.Select(x => x.SantaId).ToList();
+
+                Assert.AreEqual(pairings1.Select(x => x.OriginGroupId).First(), 1);
+                Assert.AreEqual(pairings2.Select(x => x.OriginGroupId).First(), 2);
```
Keboo: In these cases you probably want to verify that all object in the collection have the group id, rather than just the first.
```suggestion
                Assert.IsTrue(pairings2.All(x => x.OriginalGroupId == 2));
```

SecretSanta/test/SecretSanta.Api.Tests/Controllers/PairingControllerTests.cs
```diff
@@ -0,0 +1,95 @@
...
+
+            var service = new Mock<IPairingService>(MockBehavior.Strict);
+            service.Setup(x => x.GeneratePairings(groupId))
+                .ReturnsAsync(Pairings)
+                .Verifiable();
+
+            var controller = new PairingsController(service.Object, Mapper.Instance);
+
+            var result = await controller.Post(groupId) as CreatedResult;
+            List<PairingViewModel> resultPairings = result.Value as List<PairingViewModel>;
```
Keboo: After casting an object using `as` you want to make sure you check it for null before using it. If you expect the cast to succeed, use a direct cast rather than an `as`

SecretSanta/test/SecretSanta.Api.Tests/Controllers/PairingControllerTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        private CustomWebApplicationFactory<Startup> Factory { get; set; }
+        private List<Pairing> Pairings { get; set; }
+
+        [TestInitialize]
+        public void CreateWebFactory()
+        {
+            Factory = new CustomWebApplicationFactory<Startup>();
+        }
+
+        private Task<List<Pairing>> GetPairings(List<int> userIds, int groupId)
```
Keboo: Why is this method returning a Task when it is not doing any async work? It could simply be a regular method.

SecretSanta/test/SecretSanta.Api.Tests/TestableGiftService.cs
COsborn2: Good to know that wrapping is the correct terminology. Thanks!

PR #91, opened 2019-02-12
Assignment5

grantdwoods:


SecretSanta/src/SecretSanta.Domain/Services/RandomService.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            return Random.Next();
+        }
+
+        public int Next(int maxRange)
+        {
+            return Random.Next(maxRange);
+        }
+    }
+}
```
BHoward85: This is an interesting solution to the problem for the randomize the pairings.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,90 @@
...
+                GroupId = groupId
+            });
+
+            return pairings;
+        }
+
+        public async Task<List<Pairing>> GetPairingsByGroupId(int groupId)
+        {
+            return await DbContext.Pairings?.Where(x => x.GroupId == groupId)?.ToListAsync();
+        }
```
BHoward85: Good use of the null propagator for finding the pairings list for a given group.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,90 @@
...
+            indexRange--;
+            lock (RandomKey)
+            {
+                for(int i = 0; i < userIds.Count; i++, indexRange--)
+                {
+                    index = Random.Next(indexRange);
+
+                    randomIndices.Add(indices[index]);
+                    indices.Remove(indices[index]);
+                }
```
BHoward85: This method of randomizing the pairings seem to take a number of steps to do, I'm sure that it works, it just seem to a bit much.

SecretSanta/src/SecretSanta.Api/ViewModels/PairingViewModel.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public class PairingViewModel
+    {
+        public int Id { get; set; }
+        public int GroupId { get; set; }
+        [Required]
+        public int SantaId { get; set; }
+        [Required]
+        public int RecipientId { get; set; }
+    }
```
BHoward85: Good idea to mark the Santa Id and Recipient Id to be required for the building of the pairing.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
grantdwoods: I totally agree, the way everything is set up isn't very intuitive. Writing tests for this was kinda hard... I'll try to be better at refactoring later on!

SecretSanta/src/SecretSanta.Api/Controllers/PairingController.cs
```diff
@@ -0,0 +1,59 @@
...
+namespace SecretSanta.Api.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class PairingController : ControllerBase
+    {
+        private IPairingService PairingService { get; }
+        private IMapper Mapper { get; }
+
+        public PairingController(IPairingService pairingService, IMapper mapper)
```
Keboo: I would add null checks to the incoming parameters.

SecretSanta/src/SecretSanta.Api/SecretSanta.Api.csproj
```diff
@@ -5,6 +5,11 @@
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
Keboo: This is good (and the default way Visual Studio sets this property if you do it in the GUI), but this only turns on Warnings as Errors when you are in Debug mode. If you move `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` into the PropertyGroup above (the one without a Condition), it will apply to all configurations.

SecretSanta/src/SecretSanta.Domain/Services/GroupService.cs
```diff
@@ -16,47 +17,48 @@ public GroupService(ApplicationDbContext dbContext)
...
         }
 
-        public List<User> GetUsers(int groupId)
+        public async Task<List<User>> GetUsers(int groupId)
         {
-            return DbContext.Groups
-                .Where(x => x.Id == groupId)
-                .SelectMany(x => x.GroupUsers)
+            Group group = await DbContext.Groups.SingleOrDefaultAsync(x => x.Id == groupId);
+            List<User> users = group.GroupUsers
```
Keboo: `.SingleOrDefaultAsync` may return `null` if the group is not found. You should check for that prior to using the `group`.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SecretSanta.Domain.Services
+{
+    public class PairingService : IPairingService {
+        private IRandomService Random { get; }
+        private readonly object RandomKey = new object();
```
Keboo: Typically fields are expected to be named with a leading underscore.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
Keboo: Though in the `null` case this code throws a null reference exception.
Lets consider the case where `DbContext.Pairings` is in fact null.
```C#
return await (null)?.Where(x => x.GroupId == groupId).ToListAsync();
```
The null propagation short circuits the rest of the call and we are left with:
```C#
return await (Task)null;
```
Awaiting on a null task causes a null reference exception to be thrown. In general if you wanted to handle this you would either do the null check prior to the await, or use the null coalescing operator to handle the null case:
```C#
return await DbContext.Pairings?.Where(x => x.GroupId == groupId).ToListAsync() ?? Task.FromResult<List<Pairing>>(null);
```

SecretSanta/test/SecretSanta.Api.Tests/Controllers/GroupControllerTests.cs
```diff
@@ -126,41 +127,41 @@ public void UpdateGroup_ReturnsUpdatedGroup()
...
         }
 
         [TestMethod]
-        public void DeleteGroup_ReturnsNotFoundWhenTheGroupFailsToDelete()
+        public async Task DeleteGroup_ReturnsNotFoundWhenTheGroupFailsToDelete()
         {
             var service = new Mock<IGroupService>();
             service.Setup(x => x.DeleteGroup(2))
-                .Returns(false)
+                .Returns(Task.FromResult(false))
```
Keboo: There is also a `.ReturnsAsync` extension method that you can leverage to save having to type out `Task.FromResult` every time.
```suggestion
                .ReturnsAsync(false)
```

SecretSanta/test/SecretSanta.Api.Tests/Controllers/PairingControllerTests.cs
```diff
@@ -0,0 +1,148 @@
...
+using System.Threading.Tasks;
+
+namespace SecretSanta.Api.Tests.Controllers
+{
+    [TestClass]
+    public class PairingControllerTests
+    {
+        private AutoMocker Mocker { get; set; }
+        private Mock<IPairingService> MockPairingService { get; set; }
+        private Mock<IMapper> MockMapper { get; set; }
```
Keboo: This appears to be un-used. Given that you are explicitly telling AutoMocker to use the real IMapper, I would suspect that this is not needed.

SecretSanta/src/SecretSanta.Domain/Services/Interfaces/IRandomService.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SecretSanta.Domain.Services
+{
+    public interface IRandomService
+    {
+        int Next();
```
Keboo: This appears to be un-used

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            for(int i = 0; i < userIds.Count; i++)
+            {
+                indices.Add(i);
+            }
+
+            int indexRange = indices.Count;
+            indexRange--;
+
+            lock (RandomKey)
```
Keboo: This is fine and works, but there are some code smells here.
1. You are locking around much large sections of code than you need to. Only the calls to the `IRandomService` need the lock, but you are holding it for a lot longer than that.
2. There is some implicit coupling here. The _implementation_ of `IRandomService` is what actually needs the lock. There is nothing in this code here that actually needs to be locked on. This also could pose a threading problem if you were to use the `IRandomService` in another service. Because the service is shared, but the lock object is restricted to this class two separate threads could end up acting on it. A better solution would be to move the `RandomKey` field into the `RandomService` and lock on it in there.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
grantdwoods: Thank you for clarifying, still trying to get a grasp on null coalescing/null propagation!

SecretSanta/test/SecretSanta.Api.Tests/Controllers/PairingControllerTests.cs
grantdwoods: Ahh yes, I had his in place before you told me about the .Use method and missed removing it!

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
grantdwoods: I recall you mentioning trying to limit the work inside the lock, and entirely didn't think about offering synchronization in the service itself... Thanks.

PR #92, opened 2019-02-14
Assignment5 - Brad Howard

BHoward85:
All tests, functions, classes are built to spec
Pairing services, Pairing Controller, Pairing View Model, and async switch over for all services, controllers, and unit tests.

### DIFF
--- a/SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
+++ b/SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
@@ -1,0 +1,94 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using SecretSanta.Domain.Models;
+using SecretSanta.Domain.Services;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SecretSanta.Domain.Tests.Services
+{
+    [TestClass]
+    public class PairingServiceTests : DatabaseServiceTests
+    {
+        [TestMethod]
+        public async Task MakePairs_Success()
+        {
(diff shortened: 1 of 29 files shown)


SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,94 @@
...
+                while (i != 5)
+                {
+                    await groupService.AddUserToGroup(1, i);
+                    i++;
+                }
+
+                bool result = await pairingService.GenerateAllPairs(1);
+
+                Assert.IsTrue(result);
+                Assert.AreEqual(4, context.Pairings.Local.Count);
```
Keboo: It would have been nice to have some checks to verify that the pairings that were created were valid according to the assignment rules.

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            using (var context = new ApplicationDbContext(Options))
+            {
+                GroupService groupService = new GroupService(context);
+                UserService userService = new UserService(context);
+                PairingService pairingService = new PairingService(context);
+
+                bool result = await pairingService.GenerateAllPairs(1);
+
+                Assert.IsFalse(result);
+                Assert.AreEqual(0, context.Pairings.Local.Count);
```
Keboo: In general you don't want to go against the `.Local` because it only shows you the current database context's view, not always what is in the database. Usually you would do this check calling `.CountAsync()` to query the database.
```suggestion
                Assert.AreEqual(0, await context.Pairings.CountAsync());
```

PR #94, opened 2019-02-15
Assignment5

COsborn2:


SecretSanta/src/SecretSanta.Api/Controllers/PairingController.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            if (groupId <= 0)
+            {
+                return NotFound();
+            }
+
+            List<PairingViewModel> pairings = (await PairingService.GeneratePairings(groupId))
+                .Select(x => Mapper.Map<PairingViewModel>(x)).ToList();
+
+            return Created($"api/Pairing/{groupId}", pairings);
```
Keboo: I am not sure this is really the right status code to return. The url passed with a 201 is expected to be the url that can be used to retrieved the created resource. However this endpoint does not exist. Since it would try to query it with an HTTP GET. As far as I can tell this route will not map to any of your controller methods.

SecretSanta/src/SecretSanta.Api/SecretSanta.Api.csproj
```diff
@@ -1,10 +1,18 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.2</TargetFramework>
     <AspNetCoreHostingModel>InProcess</AspNetCoreHostingModel>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
```
Keboo: I realize this is what the IDE GUI does when you enable this setting. However, if you move `<TreatWarningsAsErrors>true</TreatWarningsAsErrors>` into the PropertyGroup above (the one without a Condition), it will apply to all configurations.

SecretSanta/src/SecretSanta.Domain/Services/BetterRandom.cs
```diff
@@ -0,0 +1,20 @@
+using System;
+using SecretSanta.Domain.Services.Interfaces;
+
+namespace SecretSanta.Domain.Services
+{
+    public class BetterRandom : IRandom
+    {
+        private static readonly Random RandomInstance = new Random();
```
Keboo: You really don't want this to be static. This means that all threads are forced to block just to get a random value. It would have been better to make this an instance field and let the DI container determine if this should be a singleton or not.

SecretSanta/src/SecretSanta.Domain/Services/Interfaces/IRandom.cs
```diff
@@ -0,0 +1,8 @@
+namespace SecretSanta.Domain.Services.Interfaces
+{
+    public interface IRandom
+    {
+        int Next();
+        int Next(int min, int max);
```
Keboo: This method appears to be un-used.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using SecretSanta.Domain.Models;
+using SecretSanta.Domain.Services.Interfaces;
+
+namespace SecretSanta.Domain.Services
+{
+    public class PairingService : IPairingService
+    {
+        public PairingService(ApplicationDbContext dbContext, IRandom random = null)
```
Keboo: You don't want to use an optional parameter here. You wanted to pass in a mock/testable `IRandom` instance to make your unit tests deterministic.

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,217 @@
...
+        public async Task PairingService_GeneratePairings_Success()
+        {
+            using (var context = new ApplicationDbContext(Options))
+            {
+                var temp = await SeedDatabase(3);
+                var pairingService = new PairingService(context);
+                var result = await pairingService.GeneratePairings(temp.groupId);
+                
+                Assert.AreEqual(3, result.Count);
+                // Ensure that a loop is created
```
Keboo: This comment is a bit misleading as it doesn't actually verify a loop.

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,217 @@
...
+                Assert.AreEqual(result[0].SantaId, result[2].RecipientId);
+            }
+        }
+       
+        [TestMethod]
+        public async Task PairingService_PairingsAreRandom()
+        {
+            using (var context = new ApplicationDbContext(Options))
+            {
+                var temp = await SeedDatabase(5);
```
Keboo: This would have been a good time to avoid `var`

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        [TestMethod]
+        public async Task PairingService_NotMultipleLoops()
+        {
+            using (var context = new ApplicationDbContext(Options))
+            {
+                int userIds = 12;
+                var temp = await SeedDatabase(userIds);
+                var pairingService = new PairingService(context);
+                var result = await pairingService.GeneratePairings(temp.groupId); // let this be truly random
```
Keboo: I would suggest if the goal for this test is actual random, than to have the test pass in an instance of `BetterRandom`

SecretSanta/src/SecretSanta.Api/SecretSanta.Api.csproj
COsborn2: Huh. Any reason why Rider/VS invoke the line by line approach? Also are there other configurations other than Release and Debug that the line by line approach won't hit? I'll move that up in my next Assignment. Thanks.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
COsborn2: I was trying to have an optional parameter so that in a production environment a "RealRandom" would be used. Could I make this parameter required, and still have that production Random passed in?

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
COsborn2: Ugh. I think that was Rider again. Im turning that off right now. Thanks.

SecretSanta/src/SecretSanta.Api/SecretSanta.Api.csproj
Keboo: The reason is somewhat historical. In visual studio the reason is because the GUI was build to interact with a csproj file that was always configuration based since the old csproj format this was largely always true. However the new csproj format is designed to be friendly and easy for developers to edit directly. Unfortunately the GUIs have not caught up with that.
As for other configurations, you can technically declare any configuration name you want. Debug and Release are just the common built-in ones, but there is nothing stopping you from using your own.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
Keboo: Of course. If you register it with your DI container, it will pass in the real `IRandom` instance when it creates this service. Then for your unit testing you can control which instance you want to pass in.

COsborn2: Oh, right. Sweet!

PR #97, opened 2019-02-15
Assignment 5

joeriddles:


SecretSanta/src/SecretSanta.Api/Controllers/GiftsController.cs
```diff
@@ -33,7 +33,7 @@ public IActionResult GetGiftForUser(int userId)
             {
                 return NotFound();
             }
-            List<Gift> databaseUsers = GiftService.GetGiftsForUser(userId);
+            List<Gift> databaseUsers = GiftService.GetGiftsForUser(userId).Result;
```
Keboo: This is not correct. You want to convert this to a async method and await the result. 
```C#
public IActionResult GetGiftForUser(int userId)
{
    if (userId <= 0)
    {
        return NotFound();
    }
    List<Gift> databaseUsers = await GiftService.GetGiftsForUser(userId);
```

SecretSanta/src/SecretSanta.Api/Controllers/PairingController.cs
```diff
@@ -0,0 +1,51 @@
...
+namespace SecretSanta.Api.Controllers
+{
+	[Route("api/[controller]")]
+	[ApiController]
+	public class PairingController : ControllerBase
+	{
+		private IPairingService PairingService { get; }
+		private IMapper Mapper { get; }
+
+		public PairingController(IPairingService pairingService, IMapper mapper)
```
Keboo: I would add null checks to these parameters.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/GroupControllerTests.cs
```diff
@@ -69,7 +69,7 @@ public void CreateGroup_ReturnsCreatedGroup()
             };
             var service = new Mock<IGroupService>();
             service.Setup(x => x.AddGroup(It.Is<Group>(g => g.Name == group.Name)))
-                .Returns(new Group
+                .ReturnsAsync(new Group
```
Keboo: This test is failing because the GroupsController was not updated to use the new async methods.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.EntityFrameworkCore;
+using SecretSanta.Domain.Models;
+using SecretSanta.Domain.Services.Interfaces;
+
+namespace SecretSanta.Domain.Services
+{
+	public class PairingService : IPairingService
+	{
+		private ApplicationDbContext DbContext { get; }
+		private SemaphoreSlim mutex = new SemaphoreSlim(0, 1);
```
Keboo: Close. You want to set the initial count to 1, not 0. This initial count is how many entrances the semaphore will give out.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,109 @@
...
+				int value = list[k];
+				list[k] = list[n];
+				list[n] = value;
+			}
+		}
+
+		public bool CheckForValidPairings(List<int> A, List<int> B)
+		{
+			if (A is null) throw new ArgumentNullException(nameof(A));
+			if (B == null) throw new ArgumentNullException(nameof(B));
```
Keboo: I would pick one way of doing a null check and stick with it.

PR #98, opened 2019-02-15
Assignment 5

Isaac-Bliss:
please check the changes to IUserService, UserService, UserController, and GiftServiceTests and see if I am converting correctly. Thanks

### DIFF
--- a/SecretSanta/src/SecretSanta.Api/Controllers/PairingController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/PairingController.cs
@@ -1,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using AutoMapper;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using SecretSanta.Domain.Services.Interfaces;
+
+namespace SecretSanta.Api.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class PairingController : ControllerBase
+    {
+        private IPairingService PairingService {get;set;}
+        private IMapper Mapper { get; set; }
+
+        public PairingController(IPairingService pairingService, IMapper mapper)
+        {
+            PairingService = pairingService;
+            Mapper = mapper;
+        }
+
+        //Post api/pairing/#
+        [HttpPost("{groupId}")]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesDefaultResponseType]
+        [ProducesResponseType(StatusCodes.Status201Created)]
+        public async Task<IActionResult> generateUserPairings(int groupId)
+        {
+            if (groupId <= 0)
+            {
+                return BadRequest();
+            }
+            var pairings = await PairingService.GenerateUserPairings(groupId);
+
+            return Created($"/pairing/{groupId}", pairings.Select(p => Mapper.Map<PairingViewModel>(p)).ToList());
+        }
+
+    }
+}
--- a/SecretSanta/src/SecretSanta.Api/Controllers/PairingController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/PairingController.cs
@@ -27,7 +27,7 @@ namespace SecretSanta.Api.Controllers
         [ProducesResponseType(StatusCodes.Status400BadRequest)]
         [ProducesDefaultResponseType]
         [ProducesResponseType(StatusCodes.Status201Created)]
-        public async Task<IActionResult> generateUserPairings(int groupId)
+        public async Task<IActionResult> GenerateUserPairings(int groupId)
         {
             if (groupId <= 0)
             {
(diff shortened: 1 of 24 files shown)


SecretSanta/test/SecretSanta.Domain.Tests/Services/GiftServiceTests.cs
```diff
@@ -67,7 +68,7 @@ public void UpdateGift()
                 GiftService giftService = new GiftService(context);
                 UserService userService = new UserService(context);
 
-                var users = userService.FetchAll();
+                var users = await userService.FetchAll();
                 var gifts = giftService.GetGiftsForUser(users[0].Id);
```
RichardTeller: If GetGiftsForUser is an asynchronous method (which it should be), you should await the call to it here. (you wouldnt want to continue in this method, if the call to GetGiftsForUser hasnt finished)

SecretSanta/test/SecretSanta.Domain.Tests/Services/GiftServiceTests.cs
```diff
@@ -81,7 +82,7 @@ public void UpdateGift()
...
```
RichardTeller: Same thing as previous comment

SecretSanta/src/SecretSanta.Api/Controllers/PairingController.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace SecretSanta.Api.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class PairingController : ControllerBase
+    {
+        private IPairingService PairingService {get;set;}
+        private IMapper Mapper { get; set; }
+
+        public PairingController(IPairingService pairingService, IMapper mapper)
```
Keboo: I would add null checks to these parameters.

SecretSanta/src/SecretSanta.Api/Controllers/PairingController.cs
```diff
@@ -0,0 +1,42 @@
...
+using SecretSanta.Domain.Services.Interfaces;
+
+namespace SecretSanta.Api.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class PairingController : ControllerBase
+    {
+        private IPairingService PairingService {get;set;}
+        private IMapper Mapper { get; set; }
```
Keboo: Because these properties are only set int he constructor they can be made readonly.
```suggestion
        private IMapper Mapper { get; }
```

SecretSanta/src/SecretSanta.Api/Controllers/PairingController.cs
```diff
@@ -0,0 +1,42 @@
...
+        [ProducesResponseType(StatusCodes.Status201Created)]
+        public async Task<IActionResult> GenerateUserPairings(int groupId)
+        {
+            if (groupId <= 0)
+            {
+                return BadRequest();
+            }
+            var pairings = await PairingService.GenerateUserPairings(groupId);
+
+            return Created($"/pairing/{groupId}", pairings.Select(p => Mapper.Map<PairingViewModel>(p)).ToList());
```
Keboo: I don't believe this is the correct HTTP result to return here. On a 201 (Created) request the url returned is suppose to indicate the URL that the caller can use to retrieve the created object. However there does not appear to be a controller method with a route that would match here.

SecretSanta/src/SecretSanta.Api/Controllers/PairingViewModel.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace SecretSanta.Api.Controllers
+{
+    internal class PairingViewModel
+    {
+    }
```
Keboo: Though not required for this assignment, it seems odd to have this view model and have it be empty. If you are not going to map any properties it might be just as well to omit this class.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Threading.Tasks;
+
+namespace SecretSanta.Domain.Services
+{
+    public class PairingService : IPairingService
+    {
+        private ApplicationDbContext DbContext {get;}
+        private IRandomService RandomService { get; }
+
+        public PairingService(ApplicationDbContext db, IRandomService randomService = null)
```
Keboo: You actually do not want to use an optional parameter here. Because you have registered your `RandomService` with the DI container (in the Startup.cs), it will automatically populate when it is created. For unit testing you want to explicitly pass in the instance that you want to test with either the `RandomService` or a testable implementat/mock.

SecretSanta/src/SecretSanta.Domain/Services/Interfaces/IRandomService.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Domain.Services.Interfaces
+{
+    public interface IRandomService
+    {
+        int Next();
+        int Next(int max);
```
Keboo: This method appears to be un-used

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+        }
+
+        [TestMethod]
+        public void MyTestMethod()
+        {
+
+        }*/
+    }
```
Keboo: Would have been nice to see this get implemented.

PR #100, opened 2019-02-15
Assignment5

RichardTeller:
Yet to complete any of the Pairing stuff

Update: Finished all pairing related tasks

### DIFF
--- a/SecretSanta/src/SecretSanta.Api/Controllers/GiftsController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/GiftsController.cs
@@ -27,13 +27,13 @@ namespace SecretSanta.Api.Controllers
 
         // GET api/Gift/5
         [HttpGet("{userId}")]
-        public IActionResult GetGiftForUser(int userId)
+        public async Task<IActionResult> GetGiftForUser(int userId)
         {
             if (userId <= 0)
             {
                 return NotFound();
             }
-            List<Gift> databaseUsers = GiftService.GetGiftsForUser(userId);
+            List<Gift> databaseUsers = await GiftService.GetGiftsForUser(userId);
 
             return Ok(databaseUsers.Select(x => Mapper.Map<GiftViewModel>(x)).ToList());
         }
--- a/SecretSanta/src/SecretSanta.Api/Controllers/PairingsController.cs
+++ b/SecretSanta/src/SecretSanta.Api/Controllers/PairingsController.cs
@@ -1,0 +1,50 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using AutoMapper;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using SecretSanta.Api.ViewModels;
+using SecretSanta.Domain.Models;
+using SecretSanta.Domain.Services.Interfaces;
+
+// For more information on enabling Web API for empty projects, visit https://go.microsoft.com/fwlink/?LinkID=397860
+
+namespace SecretSanta.Api.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class PairingsController : ControllerBase
+    {
+        private IPairingService PairingService { get; }
+        private IMapper Mapper { get; }
+
+        public PairingsController(IPairingService pairingService, IMapper mapper)
+        {
+            PairingService = pairingService;
+            Mapper = mapper;
+        }
+
+        // POST api/Pairings
+        [HttpPost]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(string), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(typeof(string), StatusCodes.Status409Conflict)]
+        [ProducesDefaultResponseType]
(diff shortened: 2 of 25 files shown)


SecretSanta/src/SecretSanta.Api/Controllers/GiftsController.cs
```diff
@@ -27,13 +27,13 @@ public GiftsController(IGiftService giftService, IMapper mapper)
 
         // GET api/Gift/5
         [HttpGet("{userId}")]
-        public IActionResult GetGiftForUser(int userId)
+        public async Task<IActionResult> GetGiftForUser(int userId)
```
ConnerV42: Good conversion, you could also add the produces attribute, and the producesresponsetype attribute to the top of these controller methods, just so that swagger will have more data to show you when using the API.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/GroupControllerTests.cs
```diff
@@ -32,13 +33,13 @@ public void GetAllGroups_ReturnsGroups()
 
             var service = new Mock<IGroupService>();
             service.Setup(x => x.FetchAll())
-                .Returns(new List<Group> { group1, group2 })
+                .Returns(Task.FromResult(new List<Group> { group1, group2 }))
```
ConnerV42: This is a good way to do it, but you could also just use ReturnsAsync(), which doesn't need require you to also use Task.FromResult

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,126 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SecretSanta.Domain.Tests.Services
+{
+    [TestClass]
+    public class PairingServiceTests : DatabaseServiceTests
+    {
+        [TestInitialize]
+        public async Task Initialize()
```
ConnerV42: I wonder if it is necessary to return a Task on a [TestInitialize] method. Are you doing this just to avoid an async void method signature?

SecretSanta/src/SecretSanta.Api/Controllers/PairingsController.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            if (groupId < 1)
+            {
+                return BadRequest("A group id must be specified");
+            }
+
+            List<Pairing> result = await PairingService.GeneratePairings(groupId);
+            if (result.Count == 0) return BadRequest();
+
+            return Created($"/pairing/{groupId}", result.Select(p => Mapper.Map<PairingViewModel>(p)).ToList());
```
Keboo: I am not sure this is the correct HTTP response. Typically the HTTP 201 (Created) response contains a URL where the caller can go to retrieve the newly created resource. However I do not see a controller method with a route that matches this url.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,62 @@
...
+            }
+
+            pairings.Add(new Pairing
+            {
+                SantaId = userIds.Last(),
+                RecipientId = userIds.First()
+            });
+
+            return pairings;
+        }
```
Keboo: Though this works the assignment called for the pairing sto be randomized.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/PairingsControllerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            Assert.IsNotNull(resultValue);
+            Assert.AreEqual(3, resultValue.Count);
+            Assert.AreEqual(1, resultValue[0].Id);
+            Assert.AreEqual(1, resultValue[0].SantaId);
+            Assert.AreEqual(2, resultValue[0].RecipientId);
+
+            Assert.AreEqual(3, resultValue[2].Id);
+            Assert.AreEqual(3, resultValue[2].SantaId);
+            Assert.AreEqual(1, resultValue[2].RecipientId);
+            service.VerifyAll();
```
Keboo: Would have been good to verify the contents of the last pairing returned.

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
Keboo: Yes since the `Initialize()` method contains async calls it is necessary to be `async Task`

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,126 @@
...
+                {
+                    if (!recipientIds.Contains(pairing.RecipientId)) recipientIds.Add(pairing.RecipientId);
+                }
+
+                Assert.AreEqual(pairings.Count, recipientIds.Count);
+            }
+        }
+
+        [TestMethod]
+        public async Task GeneratePairings_NoRecipientIsTheirOwnSanta()
```
Keboo: Great job translating the requirements into unit tests!

PR #101, opened 2019-02-17
Assignment 5 - Bryan Caesar

bcaesar42:


SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            return true;
+        }
+
+        private List<Pairing> GetPairings(List<int> userIds)
+        {
+            int index;
+
+            Random random = new Random();
+            index = random.Next(userIds.Count);
```
Keboo: It looks like you started to randomize the pairings but this appears un-used.

PR #102, opened 2019-02-17
Assignment5

raymondshiner:


SecretSanta/src/SecretSanta.Api/Controllers/PairingsController.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace SecretSanta.Api.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class PairingsController : ControllerBase
+    {
+        private IPairingService PairingService { get; set; }
+        private IMapper Mapper { get; set; }
+
+        public PairingsController(IPairingService service, IMapper mapper)
```
Keboo: I would add null checks to these parameters.

SecretSanta/src/SecretSanta.Api/Controllers/PairingsController.cs
```diff
@@ -0,0 +1,50 @@
...
+using SecretSanta.Domain.Services.Interfaces;
+
+namespace SecretSanta.Api.Controllers
+{
+    [Route("api/[controller]")]
+    [ApiController]
+    public class PairingsController : ControllerBase
+    {
+        private IPairingService PairingService { get; set; }
+        private IMapper Mapper { get; set; }
```
Keboo: Because these properties are only set in the constructor they can be made readonly.
```suggestion
        private IMapper Mapper { get; }
```

SecretSanta/test/SecretSanta.Api.Tests/Controllers/PairingControllerTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            var result = await controller.GeneratePairings(-1);
+
+            Assert.IsTrue(result is BadRequestResult);
+        }
+
+        [TestMethod]
+        public async Task GeneratePairings_PairingsListIsNull_ReturnsBadRequest()
+        {
+            var service = new TestablePairingService();
+            service.GeneratePairings_ToReturn = null;
```
Keboo: Though it is nice to be explicit, in this case I don't think it is necessary to set a property to its default value. At most a simple comment would suffice.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,76 @@
...
+            var random = new Random();
+            int n = pairings.Count;
+            while (n > 1)
+            {
+                n--;
+                int randomIndex = random.Next(n + 1);
+                Pairing temp = pairings[randomIndex];
+                pairings[randomIndex] = pairings[n];
+                pairings[n] = temp;
+            }
```
Keboo: This is not quite what was intended when I said the "pairings should be randomized." Because these are just going into a database it really doesn't matter the order that they are inserted. What actually matters is randomizing the who is the Santa and who is the Recipient on each pairing.

PR #103, opened 2019-02-18
Assignment5

wpcwilliams:
Working on tests...

### DIFF
--- a/SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
+++ b/SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
@@ -49,19 +49,19 @@ namespace SecretSanta.Domain.Services
             int i = 0;
             foreach (int userId in userIds)
             {
-                int idRecipient = userId;
                 int indexRecipient = i;
 
                 //no person can be their own santa && each recipient has only one santa
-                while (i == indexRecipient && !assignedRecipients.Contains(idRecipient) )
+                while (i == indexRecipient || assignedRecipients.Contains(userIds[indexRecipient]) )
                     indexRecipient = GetIndexLocked() % userIds.Count();
 
-                assignedRecipients.Add( userIds[indexRecipient] );
+                int recipientId = userIds[indexRecipient];
+                assignedRecipients.Add(recipientId);
 
                 var pairing = new Pairing
                 {
                     SantaId = userId,
-                    RecipientId = userIds[indexRecipient]
+                    RecipientId = recipientId
                 };
                 pairings.Add(pairing);
                 i++;
--- a/SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
+++ b/SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
@@ -1,0 +1,77 @@
+﻿using Microsoft.EntityFrameworkCore;
+using SecretSanta.Domain.Models;
+using SecretSanta.Domain.Services.Interfaces;
+using System;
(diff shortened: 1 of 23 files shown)


SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,82 @@
...
+            List<int> userIds = group?.GroupUsers?.Select(x => x.UserId).ToList();
+            if(userIds == null || userIds.Count < 2)
+            {
+                return null;
+            }
+
+            List<Pairing> pairings = await Task.Run( () => GetPairings(userIds));
+            if(pairings == null)
+                
+            await DbContext.Pairings.AddRangeAsync(pairings);
```
Keboo: There is a bug here. This statement is technically _inside_ of the `if(pairings == null)` above. So the pairings are never actually added to the database. Because the unit tests only check the returned pairings and not the values in the database this bug is not caught. There are also no unit tests that would cause `GetPairings` to return `null` which would have also caught this bug.

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,59 @@
...
+
+                UserService uService = new UserService(context);
+                User user1 = new User { FirstName = "fname1", LastName = "lname1" };
+                User user2 = new User { FirstName = "fname2", LastName = "lname2" };
+                User user3 = new User { FirstName = "fname3", LastName = "lname3" };
+                User user4 = new User { FirstName = "fname4", LastName = "lname4" };
+                uService.AddUser(user1);
+                uService.AddUser(user2);
+                uService.AddUser(user3);
+                uService.AddUser(user4);
```
Keboo: These are async method calls and should have been awaited.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/PairingControllerTests.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        [TestMethod]
+        public async Task GeneratePairings_GroupDoesNotContainMembers()
+        {
+            //group doesn't contain members
+
+
+
+            Assert.AreEqual(0, 0);
+        }
```
Keboo: Would have been nice to see these tests get finished.

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingController.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace SecretSanta.Domain.Tests.Services
+{
+    internal class PairingController
```
Keboo: Accidental class generation? You should not need controllers in the Domain level unit tests.

SecretSanta/test/SecretSanta.Domain.Tests/Services/PairingServiceTests.cs
```diff
@@ -0,0 +1,59 @@
...
+
+
+                PairingService pService = new PairingService(context);
+
+                List<Pairing> pairings = await pService.GeneratePairings(group.Id);
+
+
+                foreach (Pairing p in pairings)
+                {
+                    Assert.AreNotEqual(p.Recipient, p.SantaId);
```
Keboo: This is good, but it only checks one of the pairing requirements for being valid.

PR #105, opened 2019-02-22
Assignment 6

raymondshiner:
Because of multiple attempts, the name of the branch that is deploying correctly is PipeLineTester, Please consider the PipeLineTester build pipeline and associated release pipeline as my submission for assignment 6. I am recieving that 500 error, but it will be fixed when i finish fixing the code in assingment7. I sent an Email to mark@intelliTect.com with an invite to my project. Here is the link to the app service https://shiner-secretsantaapp2.azurewebsites.net/


MarkMichaelis:
Looks like it was all working Raymond.  Nice job!!

PR #122, opened 2019-02-27
Assignment7

RichardTeller:
Note I attempted adding a page for adding/removing users from groups.


SecretSanta/src/SecretSanta.Web/Views/Users/Remove.cshtml
```diff
@@ -0,0 +1,21 @@
+﻿@model UserViewModel
+@if (!string.IsNullOrWhiteSpace(ViewBag.ErrorMessage))
+{
+    <div>@ViewBag.ErrorMessage</div>
+}
+<form asp-route="Remove">
+    <div class="field">
+        <label class="label" asp-for="Id">User Id</label>
```
grantdwoods: Suggestions I made on the Users and Groups Index.cshtml files would save everyone the hassle of somehow knowing the Id's.

SecretSanta/src/SecretSanta.Web/Views/Users/Update.cshtml
```diff
@@ -0,0 +1,33 @@
+﻿@model UserViewModel
+@if (!string.IsNullOrWhiteSpace(ViewBag.ErrorMessage))
+{
+    <div>@ViewBag.ErrorMessage</div>
+}
+    <form asp-route="Update">
+        <div class="field">
+            <label class="label" asp-for="Id">User Id</label>
```
grantdwoods: Same comment as Users/Remove.cshtml.

SecretSanta/src/SecretSanta.Web/Controllers/GroupsController.cs
```diff
@@ -0,0 +1,126 @@
...
+        {
+            return View();
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> Remove(int id)
+        {
+            IActionResult result = View();
+
+            if (ModelState.IsValid)
```
grantdwoods: The ModelState check isn't really doing anything here, unless you were to:
-Check the int
-Decide it was invalid
-Set an error on the model state manually

If you were to pass a number that was <1 for instance, the ModelState.IsValid property would still be true

breaman:
Nice work overall on the assignment, this looks good.

PR #124, opened 2019-02-27
Assignment7

tdang2018:


SecretSanta/src/SecretSanta.Web/Controllers/GroupsController.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        [HttpGet]
+        public IActionResult Edit(int groupId)
+        {
+            ViewBag.GroupId = groupId;
+            return View();
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> Edit(GroupInputViewModel group, int groupId)
```
efueng: I like how you have the groupId as a param. I only had a GroupViewModel which required using AuotMapper. This way appears cleaner.

SecretSanta/src/SecretSanta.Web/Controllers/GroupsController.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            IActionResult result = View();
+
+            if (string.IsNullOrEmpty(group.Name))
+            {
+                ViewBag.ErrorMessage = "Group name cannot be empty!";
```
efueng: I like how you add an error message to the ViewBag. I'm going to have to steal this idea from you.

SecretSanta/src/SecretSanta.Web/Controllers/GroupsController.cs
```diff
@@ -0,0 +1,113 @@
...
+                {
+                    try
+                    {
+                        var secretSantaClient = new SecretSantaClient(httpClient.BaseAddress.ToString(), httpClient);
+                        await secretSantaClient.UpdateGroupAsync(groupId, group);
+                        result = RedirectToAction(nameof(Index));
+                    }
+                    catch (SwaggerException se)
+                    {
+                        ViewBag.ErrorMessage = se.Message;
```
efueng: I don't fully understand MVC, but if we're catching an exception shouldn't we also add an error to the ModelState so that it's invalidated?

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,116 @@
...
+                        ViewBag.ErrorMessage = se.Message;
+                    }
+                }
+            }
+
+            return result;
+        }
+
+        [HttpGet]
+        public IActionResult Edit(int userId)
```
efueng: Whoa, I'm not sure what you're doing here but it's way cleaner than what i'm doing. This is really slick, I'm going to look into this more and probably steal it from you :)

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            ViewBag.UserId = userId;
+            return View();
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> Edit(UserInputViewModel viewModel, int userId)
```
efueng: Again I like how you have the params set up so you can avoid using AutoMapper. But now I wonder if the data savings might make it work implementing AutoMapper. Then again an int probably doesn't add much to the overhead

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,116 @@
...
+                    }
+                    catch (SwaggerException se)
+                    {
+                        ViewBag.ErrorMessage = se.Message;
+                    }
+                }
+            }
+            else
+            {
+                ViewBag.ErrorMessage = "First name cannot be empty";
```
efueng: This is interesting. It seems like a good way to provide some more information about errors

SecretSanta/src/SecretSanta.Web/Views/Groups/Add.cshtml
```diff
@@ -0,0 +1,22 @@
+﻿@model GroupInputViewModel
+@if (!string.IsNullOrWhiteSpace(ViewBag.ErrorMessage))
```
efueng: I really like this and will have to steal it for myself!

SecretSanta/src/SecretSanta.Web/Views/Users/Edit.cshtml
```diff
@@ -0,0 +1,32 @@
+﻿@model UserInputViewModel
+@if (!string.IsNullOrWhiteSpace(ViewBag.ErrorMessage))
```
efueng: I really like how you're doing validation here. Another idea I'm going to have to borrow from you!

SecretSanta/src/SecretSanta.Web/Controllers/GroupsController.cs
tdang2018: @efueng : Thanks Eric

breaman:
Nice job on this assignment and good work on reusing the InputViewModels on the Update methods as well.

PR #125, opened 2019-02-27
Assignment7 - Brad Howard

BHoward85:
There was a mess of 4 branches for Assignment7 on my end after I did all the work on the Assignment6 and tried to fix it.

Full functions, Note that this Assignment7 states that it from Assignment7-1 is this was due to a git mess that got out of control. So I did a new branch that kept all of my Assignment6 work as I made the mistake of not checking on what branch I was working on. That is why it doesn't do the azure pipeline checks as it has conflicts.


SecretSanta/src/SecretSanta.Web/appsettings.Development.json
```diff
@@ -0,0 +1,12 @@
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Debug",
+      "System": "Information",
+      "Microsoft": "Information"
+    }
+  },
+  "SecretSantaApi": {
+    "BaseUri": "http://localhost:10452/"
```
Cwhite534: You should be using https here since its pretty easy to change and pretty standard at this point.

SecretSanta/src/SecretSanta.Web/Views/Shared/_Layout_Template.cshtml
```diff
@@ -0,0 +1,48 @@
...
+    <!-- Stylesheets -->
+    <% for (var filename in htmlWebpackPlugin.files.css) { %>
+    <link href="~<%= htmlWebpackPlugin.files.css[filename] %>" rel="stylesheet" />
+    <% } %>
+    <!-- End Stylesheets -->
+</head>
+<body>
+    <nav class="navbar" role="navigation" aria-label="main navigation">
+        <div class="navbar-brand">
+            <a class="navbar-item" asp-controller="Home">
```
Cwhite534: Im pretty sure if you dont add `asp-action="Index"` here then if you we on the user add page for example, and click this, it would get a 404 error.

SecretSanta/src/SecretSanta.Web/Views/Groups/Remove.cshtml
```diff
@@ -0,0 +1,11 @@
+﻿<h1 class="heading">Groups</h1>
+<ul>
+    @foreach (var group in ViewBag.Groups)
+    {
+        <li>
+            @group.Name
+            <a class="button is-danger" asp-action="Delete" asp-route-id="@group.Id" onclick="return confirm('Are you sure you want to delete @group.Name ?')">Delete</a>
```
Cwhite534: Once again, I really don't see the reason for moving this functionality to another view when the code is 90% the same.

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,155 @@
...
+            return result;
+        }
+
+        [HttpGet]
+        public async Task<IActionResult> Remove()
+        {
+            return await Index();
+        }
+
+        public async Task<IActionResult> Delete(int id)
```
Cwhite534: I think it would make sense for these methods to be named the same, but I suppose it is kind of personal preference.

SecretSanta/src/SecretSanta.Web/Views/Groups/Add.cshtml
```diff
@@ -0,0 +1,22 @@
+﻿@model GroupInputViewModel
+@if (!string.IsNullOrWhiteSpace(ViewBag.ErrorMessage))
+{
+    <div>@ViewBag.ErrorMessage</div>
```
Cwhite534: I like the way this adds the error to the screen. Maybe style it to with 'is-danger' to make it stand out as an error vs regular text

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,155 @@
...
+            IActionResult result = View();
+
+            if(ModelState.IsValid)
+            {
+                using (var httpClient = ClientFactory.CreateClient("SecretSantaApi"))
+                {
+                    try
+                    {
+                        int? id = viewModel.Id;
+                        var inputModel = new UserInputViewModel()
```
Cwhite534: Id recommend using AutoMapper like we did in the API.

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        private List<Pairing> BuildPairings(List<int> userIds)
+        {
+            List<Pairing> pairings = new List<Pairing>();
+            List<bool> hasPair = new List<bool>(userIds.Capacity);
+
+            Randomizer(ref userIds);
+
+            lock (_Locker)
```
Cwhite534: Not part of this assignment, but whats the reason to lock right here, if the code inside is already thread safe?

SecretSanta/src/SecretSanta.Domain/Services/PairingService.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace SecretSanta.Domain.Services
+{
+    public class PairingService : IPairingService
+    {
+        private readonly object _Locker = new object();
```
Cwhite534: This is re-created everytime the pairing service is instantiated. So each time another thread uses this service, it creates a separate locking mechanism.

PR #127, opened 2019-02-27
Assignment7-Eric Fueng

efueng:


breaman:
Nice work overall. One piece that would have been some nice polish would have been to pull back the group and user information when editing so that all information would not need to be entered again.

PR #129, opened 2019-02-27
Assignment 7 - Bryan Caesar

bcaesar42:


SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,58 @@
...
+                    catch(SwaggerException se)
+                    {
+                        ViewBag.ErrorMessage = se.Message;
+                    }
+                }
+            }
+
+            return result;
+        }
+    }
```
tdang2018: Hi Bryan. I did not see 2 functions about Update and Delete user in this file.

SecretSanta/src/SecretSanta.Web/Views/Shared/_Layout_Template.cshtml
```diff
@@ -0,0 +1,45 @@
...
+                Secret Santa
+            </a>
+        </div>
+
+        <div id="navbarBasicExample" class="navbar-menu">
+            <div class="navbar-start">
+                <a class="navbar-item" asp-controller="Users">
+                    Users
+                </a>
+                <a class="navbar-item">
```
tdang2018: I think you should add asp-controller="Groups" inside tag <a> for Groups

SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
breaman: I would agree here, I don't see the code that allows me to edit a user's information or delete them either.

breaman:
Not sure exactly what happened, but looks like there is no Groups controller that should have allowed me to create/edit/view groups. Did you have git issues again with this assignment?

PR #131, opened 2019-02-27
Assignment7

COsborn2:


SecretSanta/src/SecretSanta.Web/Controllers/UsersController.cs
```diff
@@ -0,0 +1,58 @@
...
+                    catch(SwaggerException se)
+                    {
+                        ViewBag.ErrorMessage = se.Message;
+                    }
+                }
+            }
+
+            return result;
+        }
+    }
```
tdang2018: Hi Cameron,
I think you forgot about implementing the two function Edit User and Delete User

SecretSanta/src/SecretSanta.Web/Controllers/GroupsController.cs
```diff
@@ -0,0 +1,62 @@
...
+                        result = RedirectToAction(nameof(Index));
+                    }
+                    catch (SwaggerException se)
+                    {
+                        ViewBag.ErrorMessage = se.Message;
+                    }
+                }
+            }
+
+            return result;
```
tdang2018: I think you forgot about implementing 2 methods EditGroup and DeleteGroup here.

breaman:
Nice work on this assignment. Got the error checking all handled and did a good job of using the tools that aspnet core provided.

PR #134, opened 2019-02-28
Assignment7

wpcwilliams:


bcaesar42:
I don't see your View file for Groups, but maybe I'm just overlooking it. Looks good.

SecretSanta/SecretSanta.Web/Views/Groups/Edit.cshtml
```diff
@@ -0,0 +1,20 @@
+﻿@model GroupViewModel
+
```
raymondshiner: Based on this class and your GroupsController Above. It looks like your Edit Functionality is Just change the name of the group. Have you considered trying to include some way to add and remove users (even if only by Id)?  Seems like just changing name of group isn't actually "editing" the group.

breaman:
Nice work overall. Code was cleanly written and easy to follow.

PR #135, opened 2019-03-06
Assignment8

joeriddles:


SecretSanta/src/SecretSanta.Api/Controllers/GroupsController.cs
```diff
@@ -0,0 +1,119 @@
...
+    {
+        private IGroupService GroupService { get; }
+        private IMapper Mapper { get; }
+        private ILogger Logger { get; }
+
+		public GroupsController(IGroupService groupService, IMapper mapper, ILoggerFactory loggerFactory)
+        {
+            GroupService = groupService;
+            Mapper = mapper;
+            Logger = loggerFactory.CreateLogger("GiftsController");
```
COsborn2: This should be be GroupsController.

PR #140, opened 2019-03-07
Assignment8

grantdwoods:
Pretty much just analysis stuff at this moment.


grantdwoods:
Link to Azure Application Insights: https://portal.azure.com/#@eagles.ewu.edu/resource/subscriptions/1be96e15-9a3f-408c-ba46-14fab07e631c/resourceGroups/gwassignment8/providers/microsoft.insights/components/gwAssignment8/overview

SecretSanta/src/SecretSanta.Api/Program.cs
```diff
@@ -61,7 +46,26 @@ public static void Main(string[] args)
 
         public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
             WebHost.CreateDefaultBuilder(args)
+                .UseApplicationInsights()
                 .UseStartup<Startup>()
+                .ConfigureAppConfiguration((hostingContext, config) =>
```
grantdwoods: Configuration providers:
    File < Environment < Command Line
    Order seems to match up with what was discussed in class.

From what I saw in the documentation, if a duplicate key/value is found the last one encountered takes priority. If someone bothered entering configuration in through the terminal it should take priority over anything else. Same with setting up env variables over whats in a file. I think of this like how forced Visual Studio to use our version of Node(from env/path) instead of whatever it installed on it's own.

SecretSanta/src/SecretSanta.Api/Program.cs
```diff
@@ -61,7 +46,26 @@ public static void Main(string[] args)
 
         public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
             WebHost.CreateDefaultBuilder(args)
+                .UseApplicationInsights()
```
MarkMichaelis: I don't believe this is necessary to explicitly configure in .NET Core ~2.0 or later.

PR #149, opened 2019-03-11
Assignment9-Eric Fueng

efueng:
Pipeline test


SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            deleteLink.Click();
+            Driver.SwitchTo().Alert().Accept();
+
+            //Assert
+            List<string> userNames = page.UserNames;
+            Assert.IsFalse(userNames.Contains($"{firstName} {lastName}"));
+        }
+
+        [TestMethod]
+        public void FailToDeleteUserTakesScreenshot()
```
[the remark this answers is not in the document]
efueng: Hi Tuan. This is a poorly named method from me. The code to take the screenshot is in the test cleanup method. This test just intentionally fails in order to trigger the screenshot.

SecretSanta/test/SecretSanta.Web.UITests/GroupsPageTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        public void Init()
+        {
+            Driver = new ChromeDriver(Path.GetFullPath("."));
+        }
+
+        [TestCleanup]
+        public void Cleanup()
+        {
+            Driver.Quit();
+            Driver.Dispose();
```
PandaMagnus: Quit() and Dispose() do the same thing (well, more specifically, Quit calls Dispose). I prefer using Quit() since Dispose() has specific connotations that might be misleading if you're not invoking Selenium as part of a using statement, but it's honestly 100% up to you which one you use.

SecretSanta/test/SecretSanta.Web.UITests/GroupsPageTests.cs
```diff
@@ -0,0 +1,111 @@
...
+            //uri.Host
+            //Assert
+            Assert.IsTrue(Driver.Url.EndsWith(GroupsPage.Slug));
+        }
+
+        [TestMethod]
+        public void CanNavigateToAddGroupsPage()
+        {
+            //Arrange
+            var rootUri = new Uri(RootUrl);
```
PandaMagnus: GoToUrl has an overload that takes a string, so instead of declaring a new URI, you can simply combine RootUrl and GroupsPage.Slug as a string when calling GoToUrl. Not a huge deal, but it saves you a line every time you need to navigate somewhere relative to the root path.

SecretSanta/test/SecretSanta.Web.UITests/GroupsPageTests.cs
```diff
@@ -0,0 +1,111 @@
...
+            Driver.Navigate().GoToUrl(new Uri(rootUri, GroupsPage.Slug));
+            var groupsPage = new GroupsPage(Driver);
+            
+            //Act
+            groupsPage.AddGroupButton.Click();
```
PandaMagnus: This might be because I don't have the code base on this machine, but does clicking the "Add Group" button change the URL? This test reads like you're going to https://localhost:44334/Add, doing something, then asserting that you're on https://localhost:44334/Add.

Again, I may be misunderstanding since I don't have the website handy, but if that's the case, you could probably axe this test.

SecretSanta/test/SecretSanta.Web.UITests/GroupsPageTests.cs
```diff
@@ -0,0 +1,111 @@
...
+            //Assert
+            Assert.IsTrue(Driver.Url.EndsWith(AddGroupPage.Slug));
+        }
+
+        [TestMethod]
+        public void CanAddGroups()
+        {
+            //Arrange
+            string groupName = "Group Name" + Guid.NewGuid().ToString();
+            GroupsPage page = CreateGroup(groupName);
```
PandaMagnus: I like how you've consolidated the common flow into the CreateGroup() method, but not tried to over-complicate it to handle more than just this explicit case. It's straightforward and easy to understand.

SecretSanta/test/SecretSanta.Web.UITests/TestPages/AddGroupPage.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class AddGroupPage
+    {
+        public const string Slug = GroupsPage.Slug + "/Add";
+        public IWebDriver Driver { get; }
+        public IWebElement GroupNameTextBox => Driver.FindElement(By.Id("Name"));
+        public IWebElement SubmitButton =>
+            Driver
+            .FindElements(By.CssSelector("button.is-primary"))
+            .Single(x => x.Text == "Submit");
```
PandaMagnus: Good use of LINQ here. As an alternative, if the text is fetch-able from the DOM, you could also use a CSS or XPath selector.

SecretSanta/test/SecretSanta.Web.UITests/TestPages/AddGroupPage.cs
```diff
@@ -0,0 +1,24 @@
...
+        
+        public AddGroupPage(IWebDriver driver)
+        {
+            Driver = driver ?? throw new ArgumentNullException(nameof(driver));
```
PandaMagnus: Good use of null checking!

SecretSanta/test/SecretSanta.Web.UITests/TestPages/AddUserPage.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace SecretSanta.Web.UITests.TestPages
+{
+    public class AddUserPage
+    {
+        public const string Slug = UsersPage.Slug + "/Add";
```
PandaMagnus: Even though it's technically a different page, a similarly declared URL twice, you could potentially move it up to a base class or partial class declaration.

SecretSanta/test/SecretSanta.Web.UITests/TestPages/AddUserPage.cs
```diff
@@ -0,0 +1,25 @@
...
+        public IWebDriver Driver { get; }
+        public IWebElement FirstNameTextBox => Driver.FindElement(By.Id("FirstName"));
+        public IWebElement LastNameTextBox => Driver.FindElement(By.Id("LastName"));
+        public IWebElement SubmitButton =>
+            Driver
+            .FindElements(By.CssSelector("button.is-primary"))
+            .Single(x => x.Text == "Submit");
```
PandaMagnus: Same here, it might be possible to make this part of a base or partial class and not have to re-declare this.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,143 @@
...
+                    Environment.CurrentDirectory.Substring(0, Environment.CurrentDirectory.LastIndexOf("bin"));
+
+                string fileName = $"{projectRoot}/Screenshots/{TestContext.TestName}.png";
+
+                Screenshot screenshot = ((ITakesScreenshot) Driver).GetScreenshot();
+                screenshot.SaveAsFile(fileName, ScreenshotImageFormat.Png);
+            }
+
+            Driver.Quit();
+            Driver.Dispose();
```
PandaMagnus: Same as above, you can call just one of these and Selenium should clean itself up fine (if it doesn't, there's something else going on, which happens occasionally if there's a failure outside of a test.)

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            //Arrange
+            Driver.Navigate().GoToUrl(RootUrl);
+
+
+            //Act
+            var homePage = new HomePage(Driver);
+            homePage.UsersLink.Click();
+
+            //var uri = new Uri(Driver.Url);
+            //uri.Host
```
PandaMagnus: Minor: Unless these needs to stay for a future implementation, go ahead and delete commented out code.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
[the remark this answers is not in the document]
PandaMagnus: You could move the take screenshot code out into a helper method to invoke from test cleanup on failure, and from this test explicitly. Having a test that's expected to fail can throw people unfamiliar with the code off, so it might make more sense to turn this into an Assert.IsFalse() and then call the screenshot method after.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            string lastName = "LastName" + Guid.NewGuid().ToString();
+            UsersPage page = CreateUser(firstName, lastName);
+
+            //Act
+            var deleteLink = page.GetDeleteLink($"{firstName} {lastName}");
+            deleteLink.Click();
+            Driver.SwitchTo().Alert().Accept();
+            //Assert
+            List<string> userNames = page.UserNames;
+            Assert.IsTrue(userNames.Contains($"{firstName} {lastName}"));
```
PandaMagnus: Informative assert messages will help when other people need to investigate a test. It's not required, but try to add in helpful messages for asserts that aren't easily indicative of the comparison being made (E.G., Assert.AreEqual tends to show you the expected and actual values for most test frameworks, but Assert.IsTrue tends to not, which makes it harder to investigate just based on test results.)

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,143 @@
...
+
+            var addUserPage = new AddUserPage(Driver);
+
+            addUserPage.FirstNameTextBox.SendKeys(firstName);
+            addUserPage.LastNameTextBox.SendKeys(lastName);
+            addUserPage.SubmitButton.Click();
+
+            return page;
+        }
+    }
```
PandaMagnus: This is a good start to the tests. Think about adding an edit test and some tests for doing boundary analysis on the available text boxes.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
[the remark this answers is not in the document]
efueng: Oh that's a great idea, thank you.

SecretSanta/test/SecretSanta.Web.UITests/GroupsPageTests.cs
efueng: I'm all for saving lines of code. Thank you.

SecretSanta/test/SecretSanta.Web.UITests/GroupsPageTests.cs
efueng: We were just testing whether we could successfully navigate to the Add Group page from the Group index page.

SecretSanta/test/SecretSanta.Web.UITests/GroupsPageTests.cs
efueng: I wish I could take credit, but this part was discussed in class.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,143 @@
...
+        public void Cleanup()
+        {
+            if (TestContext.CurrentTestOutcome == UnitTestOutcome.Failed)
+            {
+                string projectRoot =
+                    Environment.CurrentDirectory.Substring(0, Environment.CurrentDirectory.LastIndexOf("bin"));
+
+                string fileName = $"{projectRoot}/Screenshots/{TestContext.TestName}.png";
+
+                Screenshot screenshot = ((ITakesScreenshot) Driver).GetScreenshot();
```
Keboo: Though this works with the current driver it would be better to type check the Driver rather than directly casting. Then if you ever use a driver that doesn't support screenshots it wont cause your test to fail.
```suggestion
                if (Driver is ITakesScreenshot takeScreenshot)
                {
                    Screenshot screenshot = takeScreenshot.GetScreenshot();
```

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            Assert.IsTrue(Driver.Url.EndsWith(AddUserPage.Slug));
+        }
+
+        [TestMethod]
+        public void CanAddUsers()
+        {
+            //Arrange
+            string firstName = "FirstName";
+            string lastName = "LastName" + Guid.NewGuid().ToString();
+            UsersPage page = CreateUser(firstName, lastName);
```
Keboo: Minor thing. I would move this down under `//Act`. Since it is technically the action being tested.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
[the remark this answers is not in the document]
Keboo: I would also rename the unit test since "TakesScreenshot" is not actually part of the test.

PR #150, opened 2019-03-12
Assignment9

Cwhite534:
In collaboration with @KennethWhite.


SecretSanta/test/SecretSanta.web.UITests/FirefoxTests.cs
```diff
@@ -0,0 +1,16 @@
...
+    [TestClass]
+    public class FireFoxTests : UserPageTests
+    {
+        [TestInitialize]
+        public void Init()
+        {
+            Driver = new FirefoxDriver(Path.GetFullPath("."));
+        }
+    }
+}
```
BHoward85: Did this work for you? I had to set FF to Accept Insecure Certificates in order for it to work. Also nice way of using inheritance as a means of reducing code duplication.

Isaac-Bliss:
looks good, I like the conciseness of your unit tests

KennethWhite:
Extra Credit: In the end @Cwhite534 and I couldn't get the tests running properly for Edge, it couldn't find the link elements using the LinkText nor using css selectors and I am not sure why. Firefox I had issues with the security ceritificate as @BHoward85 noted above. I tried with Internet Explorer drivers as well but only about half the tests passed due to issues finding elements as well.

Further Casey and I struggled to find a solution to getting multiple drivers to run the same set of tests. Initially we really liked this inheritance based solution, however tonight realized if we had continued on to do Groups or Gifts page tests we would have had to duplicate the inheriting classes again. 

We would be interested in feedback or seeing your own solution for this extra credit.

SecretSanta/test/SecretSanta.web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        [TestCleanup]
+        public void Cleanup()
+        {
+            if (TestContext.CurrentTestOutcome == UnitTestOutcome.Failed)
+            {
+                SaveScreenshotOfFailure();
+            }
+
+            Driver.Quit();
+            Driver.Dispose();
```
PandaMagnus: Quit() calls Dispose(). You only need one here. I tend to use Quit() since Dispose() has specific connotations in C#, but it's really up to you which one you use.

SecretSanta/test/SecretSanta.web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,192 @@
...
+                if (!Directory.Exists(outputDirectory))
+                {
+                    Directory.CreateDirectory(outputDirectory);
+                }
+
+                var outputFilename = Path.Join(outputDirectory, $"{TestContext.TestName}.png");
+                var screenshot = ((ITakesScreenshot) Driver).GetScreenshot();
+                screenshot.SaveAsFile(outputFilename, ScreenshotImageFormat.Png);
+            }
+            catch (Exception e) // catches IO errors/ Permission errors/ anything else
```
PandaMagnus: Might want to catch specific exceptions in cases where we can feasibly retry.

SecretSanta/test/SecretSanta.web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        {
+            //Arrange
+            Driver.Navigate().GoToUrl(RootUrl);
+
+            //ACt
+            var homePage = new HomePage(Driver);
+            homePage.UsersLink.Click();
+
+            //Assert
+            Assert.IsTrue(Driver.Url.EndsWith(UsersPage.Slug));
```
PandaMagnus: Prefer to have an assert message with asserts that tend to not provide a helpful message (E.G. Assert.Equals methods for most frameworks tend to show you the expected and actual values, while Assert.True methods tend not to.)

SecretSanta/test/SecretSanta.web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,192 @@
...
+            //Arrange
+            var rootUri = new Uri(RootUrl);
+            Driver.Navigate().GoToUrl(new Uri(rootUri, UsersPage.Slug));
+            var usersPage = new UsersPage(Driver);
+            usersPage.AddUserLink.Click();
+            string userFirstName = "First Name" + Guid.NewGuid().ToString("N");
+            string userLastName = "Last Name" + Guid.NewGuid().ToString("N");
+
+            //Act
+            CreateUser(RootUrl, userFirstName, userLastName);
```
PandaMagnus: Good job extracting out the common steps to a separate method!

SecretSanta/test/SecretSanta.web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,192 @@
...
+            IWebElement editLink = usersPage.EditLink($"{userFirstName} {userLastName}");
+            string linkText = editLink.GetAttribute("href");
+            string userID = (linkText.Substring(linkText.LastIndexOf("/") + 1));
+            var editPage = new EditUserPage(Driver);
+
+            editLink.Click();
+
+            Assert.AreEqual<string>(userID, editPage.CurrentUserID);
+            Assert.AreEqual<string>(userFirstName, editPage.FirstNameTextBox.GetAttribute("value"));
+            Assert.AreEqual<string>(userLastName, editPage.LastNameTextBox.GetAttribute("value"));
```
PandaMagnus: Good use of Assert.AreEqual here.

SecretSanta/test/SecretSanta.web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,192 @@
...
+            Assert.IsTrue(users.Contains($"{userFirstName} {userLastName}"));
+        }
+
+
+        [TestMethod]
+        public void CanDeleteUser()
+        {
+            //Arrange
+            string userFirstName = "First Name" + Guid.NewGuid().ToString("N");
+            string userLastName = "Last Name" + Guid.NewGuid().ToString("N");
```
PandaMagnus: Since this has shown up a couple of times, you could abstract this out to one or two properties that gets generated on test initialize (besides having one property for first name and one for last, you could use a Dictionary<string, string> or Tuple<string, string>)

SecretSanta/test/SecretSanta.web.UITests/WebPages/UserPages/AddUserPage.cs
```diff
@@ -0,0 +1,27 @@
...
+        public IWebElement UserLastNameTextBox => Driver.FindElement(By.Id("LastName"));
+
+        public IWebElement SubmitButton => 
+            Driver
+                .FindElements(By.CssSelector("button.is-primary"))
+                .Single(x => x.Text == "Submit");
+
+        public AddUserPage(IWebDriver driver)
+        {
+            Driver = driver ?? throw new ArgumentNullException(nameof(driver));
```
PandaMagnus: Good use of null checking on this.

SecretSanta/test/SecretSanta.web.UITests/WebPages/UserPages/UsersPage.cs
```diff
@@ -0,0 +1,61 @@
...
+            ReadOnlyCollection<IWebElement> deleteLinks =
+                Driver.FindElements(By.CssSelector("a.is-danger"));
+
+           return deleteLinks.Single(x => x.GetAttribute("onclick").EndsWith($"{firstName} {lastName}?')"));
+        }
+
+        public IWebElement EditLink(string userName)
+        {
+            ReadOnlyCollection<IWebElement> userElements =
+                Driver.FindElements(By.CssSelector("h1+ul>li"));
```
PandaMagnus: Since you have this selector declared twice, you could pull it out into its own property.

SecretSanta/test/SecretSanta.web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,192 @@
...
+            page.AddUserLink.Click();
+
+            var addUserPage = new AddUserPage(Driver);
+
+            addUserPage.UserFirstNameTextBox.SendKeys(firstName);
+            addUserPage.UserLastNameTextBox.SendKeys(lastName);
+            addUserPage.SubmitButton.Click();
+            return page;
+        }
+    }
```
PandaMagnus: Good set of tests! Consider adding some boundary analysis around the available text boxes.

PandaMagnus:
@KennethWhite @Cwhite534 There's a couple of different ways to get multi-browser tests up-and-running. None of them are super straightforward, and they all have their tradeoffs. The most common methods I've seen are:

- Make the browser selection drive by a parameter on a base class or your test initialize method. Pass in that parameter via a config file on test run
- Have the browser be selected as part of a data driven test. E.G., something like here but with browser types: https://stackoverflow.com/questions/44789698/is-there-a-better-way-to-pass-dynamic-inputs-in-line-to-a-datatestmethod-i-e-h

SecretSanta/test/SecretSanta.web.UITests/SecretSanta.web.UITests.csproj
```diff
@@ -0,0 +1,20 @@
...
+        <PackageReference Include="MSTest.TestAdapter" Version="1.3.2" />
+        <PackageReference Include="MSTest.TestFramework" Version="1.3.2" />
+        <PackageReference Include="Selenium.Firefox.WebDriver" Version="0.23.0" />
+        <PackageReference Include="Selenium.WebDriver" Version="3.141.0" />
+        <PackageReference Include="Selenium.WebDriver.ChromeDriver" Version="72.0.3626.69" />
+        <PackageReference Include="Selenium.WebDriver.IEDriver" Version="3.141.5" />
+        <PackageReference Include="Selenium.WebDriver.MicrosoftDriver" Version="17.17134.0" />
+    </ItemGroup>
+
+</Project>
```
Keboo: Minor thing here. But the file name should match the casing of the project being tested. `SecretSanta.Web.UITests.csproj`

SecretSanta/test/SecretSanta.web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,192 @@
...
+                var projectDirectory = Directory.GetCurrentDirectory().Split(@"\bin").First();
+                var outputDirectory = Path.Join(projectDirectory, "FailedTests");
+
+                if (!Directory.Exists(outputDirectory))
+                {
+                    Directory.CreateDirectory(outputDirectory);
+                }
+
+                var outputFilename = Path.Join(outputDirectory, $"{TestContext.TestName}.png");
+                var screenshot = ((ITakesScreenshot) Driver).GetScreenshot();
```
Keboo: This works because the current driver supports screenshots, but what happens when the Driver doesn't implement ITakesScreenshot? Better to type check rather than simply cast.
```suggestion
                if (Driver is ITakesScreenshot takesScreenshot)
                {
                    var screenshot = takesScrrenshot.GetScreenshot();
```

SecretSanta/test/SecretSanta.web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            Driver.Quit();
+            Driver.Dispose();
+        }
+
+        private void SaveScreenshotOfFailure()
+        {
+            try
+            {
+                var projectDirectory = Directory.GetCurrentDirectory().Split(@"\bin").First();
```
Keboo: This makes a bit of an assumption on the working directory (and on the operating system being Windows because of the backslash). I would consider simply using the current directory rather than trying to look higher up.

SecretSanta/test/SecretSanta.web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,192 @@
...
+                var outputDirectory = Path.Join(projectDirectory, "FailedTests");
+
+                if (!Directory.Exists(outputDirectory))
```
Keboo: Don't do an Exists check prior to calling Create. Just call Directory.Create.

SecretSanta/test/SecretSanta.web.UITests/EdgeTests.cs
```diff
@@ -0,0 +1,17 @@
+using System.IO;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using OpenQA.Selenium.Edge;
+
+
+namespace SecretSanta.web.UITests
+{
+    //[TestClass]
+    //public class EdgeTests : UserPageTests
```
Keboo: This type of setup is exactly correct for doing tests on multiple browsers. However, getting a system with all of the right browsers installed can be an issue. For things like edge if you checks versions I suspect you probably have a newer version than what the selenium driver supports.

PR #152, opened 2019-03-13
Assignment9

grantdwoods:
Still some refactoring to do, as well as tests to verify header links and cancel link.

### DIFF
--- a/SecretSanta/test/SecretSanta.Web.UITests/BaseTestPage.cs
+++ b/SecretSanta/test/SecretSanta.Web.UITests/BaseTestPage.cs
@@ -1,0 +1,40 @@
+﻿using OpenQA.Selenium;
+using SecretSanta.Web.UITests.Pages;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Web.UITests
+{
+    public class BasePageTest
+    {
+        protected IWebDriver Driver { get; set; }
+        protected const string RootUrl = "https://localhost:44314/";
+    
+        protected GroupsPage CreateGroup(string groupName)
+        {
+            var page = CreateGroupsPage();
+            page.AddGroup.Click();
+
+            var addGroupPage = new AddGroupsPage(Driver);
+
+            addGroupPage.GroupNameTextBox.SendKeys(groupName);
+            addGroupPage.SubmitButton.Click();
+            return page;
+        }
+
+        protected GroupsPage CreateGroupsPage()
+        {
+            var rootUri = new Uri(RootUrl);
+            Driver.Navigate().GoToUrl(new Uri(rootUri, GroupsPage.Slug));
+            return new GroupsPage(Driver);
+        }
+
+        protected UsersPage CreateUsersPage()
+        {
+            var rootUri = new Uri(RootUrl);
+            Driver.Navigate().GoToUrl(new Uri(rootUri, UsersPage.Slug));
+            return new UsersPage(Driver);
+        }
+    }
+}
--- a/SecretSanta/test/SecretSanta.Web.UITests/NavigationTests.cs
+++ b/SecretSanta/test/SecretSanta.Web.UITests/NavigationTests.cs
@@ -1,0 +1,142 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using OpenQA.Selenium;
+using OpenQA.Selenium.Chrome;
+using SecretSanta.Web.UITests.Pages;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace SecretSanta.Web.UITests
+{
+    [TestClass]
+    public class NavigationTests
+    {
+        private const string RootUrl = "https://localhost:44314/";
+
+        private IWebDriver Driver { get; set; }
+
+        [TestInitialize]
+        public void Init()
+        {
+            Driver = new ChromeDriver(Path.GetFullPath("."));
+        }
+
+        [TestCleanup]
+        public void Cleanup()
+        {
+            Driver.Quit();
+            Driver.Dispose();
+        }
+        [TestMethod]
+        public void FromHomePage_ToUsersPage_ByUsersLink()
+        {
+            Driver.Navigate().GoToUrl(RootUrl);
+
+            var homePage = new HomePage(Driver);
+            homePage.UsersLink.Click();
+
+            Assert.IsTrue(Driver.Url.EndsWith(UsersPage.Slug));
+        }
+
+        [TestMethod]
+        public void FromHomePage_ToGroupsPage_BySecretSantaLink()
+        {
+            Driver.Navigate().GoToUrl(RootUrl);
+
+            var homePage = new HomePage(Driver);
+            homePage.GroupsLink.Click();
+
+            Assert.IsTrue(Driver.Url.EndsWith(GroupsPage.Slug));
+        }
+
+        [TestMethod]
+        public void FromUsersPage_ToHomePage_BySecretSantaLink()
+        {
+            Driver.Navigate().GoToUrl($"{RootUrl}{UsersPage.Slug}");
+            UsersPage usersPage = new UsersPage(Driver);
+
+            usersPage.HomePageLink.Click();
+
+            Assert.AreEqual<string>(RootUrl, Driver.Url);
+        }
+
+        [TestMethod]
+        public void FromUsersPage_ToAddUserPage_ByAddUserLink()
+        {
+            var usersPage = CreateUsersPage();
+
+            usersPage.AddUser.Click();
+
+            Assert.IsTrue(Driver.Url.EndsWith(AddUserPage.Slug));
+        }
+
+        [TestMethod]
+        public void FromUsersPage_ToGroupsPage_ByGroupsLink()
+        {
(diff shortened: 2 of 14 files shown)


SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,206 @@
...
+            var usersPage = CreateUsersPage();
+            usersPage.AddUser.Click();
+            GuidUserInfo(out string first, out string last);
+            CreateUser(first, last);
+            usersPage.GetEditLink($"{first} {last}").Click();
+            EditUserPage editPage = new EditUserPage(Driver);
+            editPage.FirstNameTextBox.Clear();
+
+            Assert.IsFalse(editPage.ErrorElement.Displayed);
+            editPage.SubmitButton.Click();
```
grantdwoods: Test fails, 500 is thrown by server. Caused by trying to send null/empty string.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -51,6 +61,17 @@ public void CanNavigateToAddUserPage()
             Assert.IsTrue(Driver.Url.EndsWith(AddUserPage.Slug));
         }
 
+        [TestMethod]
+        public void CanNavigateBackHomeFromAddUserPage()
```
grantdwoods: This test fails due to a bug in the layout template.

SecretSanta/test/SecretSanta.Web.UITests/NavigationTests.cs
```diff
@@ -0,0 +1,142 @@
...
+            Driver.Navigate().GoToUrl($"{RootUrl}{UsersPage.Slug}");
+            UsersPage usersPage = new UsersPage(Driver);
+
+            usersPage.GiftsPageLink.Click();
+
+            Assert.AreEqual<string>($"{RootUrl}{GiftsPage.Slug}", Driver.Url);
+        }
+
+        [TestMethod]
+        public void FromAddUserPage_ToHomePage_BySecretSantaLink()
```
grantdwoods: This test fails due to a bug in the layout template.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
BHoward85: I put a if null check in the create users function. So that when I send in a null is just skips the text box. You can also do this by clearing the first name box and not adding any text, this should avoid the 500 error.

grantdwoods: I appreciate you peeking in! 
However, this test failing is unearthing a bug in the system under test. The appropriate behavior I would expect is to not get an error that crashes everything. By changing my test to avoid the 500 I would effectively just be pretending like I didn't find an issue!

SecretSanta/test/SecretSanta.Web.UITests/NavigationTests.cs
grantdwoods: The same error appears from any Edit/Add page. I already know the number of tests here are unnecessary since they all go against the same template, but I'm not sure how many are sufficient in proving the bug.

SecretSanta/test/SecretSanta.Web.UITests/BaseTestPage.cs
```diff
@@ -0,0 +1,40 @@
+﻿using OpenQA.Selenium;
+using SecretSanta.Web.UITests.Pages;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SecretSanta.Web.UITests
+{
+    public class BasePageTest
```
grantdwoods: I feel like this is a bad name. Any suggestions are appreciated!

Keboo: Personally I would suggest `TestBase`. I would also then suggest moving the Create* methods into extension methods.

SecretSanta/test/SecretSanta.Web.UITests/NavigationTests.cs
Keboo: I think the number of tests is fine. Remember when writing these tests you want to approach it from an end user perspective rather than from the perspective of the developer. From the developer perspective the bug exists in a single template but from an end user perspective there are several similar looking bugs.

SecretSanta/test/SecretSanta.Web.UITests/NavigationTests.cs
```diff
@@ -0,0 +1,124 @@
...
+using SecretSanta.Web.UITests.Pages;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace SecretSanta.Web.UITests
+{
+    [TestClass]
+    public class NavigationTests : BasePageTest
```
Keboo: I like how you structured this with the overall site navigation in its own test class :+1:

SecretSanta/test/SecretSanta.Web.UITests/Pages/AddUserPage.cs
```diff
@@ -0,0 +1,30 @@
+﻿using OpenQA.Selenium;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace SecretSanta.Web.UITests.Pages
+{
+    class AddUserPage
```
Keboo: You want to explicitly specify the protection modifier on everything (I realize the default Visual Studio template omits it).

SecretSanta/test/SecretSanta.Web.UITests/Pages/EditUserPage.cs
```diff
@@ -0,0 +1,28 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace SecretSanta.Web.UITests.Pages
+{
+    class EditUserPage
+    {
+        IWebDriver Driver { get; }
```
Keboo: Again, you want to specify the protection modifier here.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,167 @@
...
+            string userFull = $"{first} {last}";
+            usersPage.GetDeleteLink(userFull).Click();
+            Driver.SwitchTo().Alert().Accept();
+
+            var users = usersPage.UserNames.ToList();
+
+            Assert.IsFalse(users.Contains(userFull));
+        }
+
+        private void GuidUserInfo(out string userFirst, out string userLast)
```
Keboo: Nice test method. As an alternate syntax you could have returned a tuple, which would have allowed callers to deconstruct the tuple into individual variables.

SecretSanta/test/SecretSanta.Web.UITests/Pages/UsersPage.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            Driver = driver ?? throw new ArgumentNullException(nameof(driver));
+        }
+        public IEnumerable<string> UserNames
+        {
+            get
+            {
+                var userListItems = Driver.FindElements(By.CssSelector("h1.title+ul>li"));
+                return userListItems.Select(x => 
+                    x.Text.Substring(0,x.Text.Length - UserLinksText.Length)).ToList();
```
Keboo: I would omit the `.ToList()` portion or change the return type to something like `IList<string>`. By generating the list here, and then returning an `IEnumerable<string>` you make the caller then call `.ToList` on something that is already a list.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,167 @@
...
+        [TestMethod]
+        public void CanAddUser()
+        {
+            GuidUserInfo(out string first, out string last);
+
+            UsersPage usersPage = CreateUser(first, last);
+
+            List<string> users = usersPage.UserNames.ToList();
+
+            Assert.IsTrue(users.Contains($"{first} {last}"));
```
Keboo: There is technically a [Contains LINQ method](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.contains?view=netcore-2.0) that you could have leveraged (since UserNames returns an `IEnumerable<string>`). This would save you from needing to construct a list.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,167 @@
...
+        [TestMethod]
+        public void CanNotCangeToEmptyFirstNameErrorAppears()
+        {
+            GuidUserInfo(out string first, out string last);
+            UsersPage usersPage = CreateUser(first, last);
+            usersPage.GetEditLink($"{first} {last}").Click();
+            EditUserPage editPage = new EditUserPage(Driver);
+            editPage.FirstNameTextBox.Clear();
+
+            Assert.IsFalse(editPage.ErrorElement.Displayed);
```
Keboo: Typically if you need to put an `Assert` call in the middle of a unit test, it is a sign that you should have two separate tests.

PR #153, opened 2019-03-13
Assignment9 - Chester Southwood

Chester-Southwood:


SecretSanta/SecretSanta.Web.UITests/GroupPageTests.cs
```diff
@@ -0,0 +1,184 @@
...
+        //public IWebElement GroupsLink => Driver.FindElement(By.CssSelector("a[href=\"/Groups\"]"));
+        public IWebElement GroupsLink => Driver.FindElement(By.LinkText("Groups"));
+
+        public HomePage(IWebDriver driver)
+        {
+            Driver = driver ?? throw new ArgumentNullException(nameof(driver));
+        }
+    }
+
+    public class GroupsPage
```
efueng: consider putting the page classes into their own files

SecretSanta/test/SecretSanta.Web.UITests/PagesTests/UsersPageTests.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            UsersPageAdd addUserPage = new UsersPageAdd(Driver);
+            addUserPage.UserFirstNameTextBox.Clear();
+            addUserPage.UserFirstNameTextBox.SendKeys(userFirstName);
+            addUserPage.UserLastNameTextBox.Clear();
+            addUserPage.UserLastNameTextBox.SendKeys(userLastName);
+            addUserPage.SubmitButton.Click();
+        }
+
+        private struct StructUser
```
Keboo: Typically it is best to avoid including things like struct, class", etc in the type name. As a simple alternative you might have considered using a tuple in place of this.

PR #155, opened 2019-03-13
Assignment9 Submission

raymondshiner:


SecretSanta/test/SecretSanta.Web.UITests/DriverPages/AddGroupsPage.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Linq;
+using OpenQA.Selenium;
+
+namespace SecretSanta.Web.UITests.DriverPages
+{
+    public class AddGroupsPage
+    {
+        public const string Path = GroupsPage.Path + "Add/";
```
BHoward85: I see that you have a number of these const strings that are called path, it seems to be has an overlap with the Slug.
Is there a reason for it?

SecretSanta/test/SecretSanta.Web.UITests/DriverPages/AddUsersPage.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public IWebElement SubmitButton =>
+            Driver
+                .FindElements(By.CssSelector("button.is-primary"))
+                .Single(x => x.Text == "Submit");
+
+        public IWebElement GetNavbarLink(string linkName) => Driver.FindElement(By.LinkText(linkName));
+
+        public IWebElement GetFirstNameRequiredNotification => Driver.FindElement(By.CssSelector("ul>li"));
+    }
```
BHoward85: This is a better way then what I did, I looked for the error inside the test.

SecretSanta/test/SecretSanta.Web.UITests/DriverPages/AddGroupsPage.cs
raymondshiner: Yes. So path is the whole path. This way over in my tests when i want to navigate somewhere like add users page from a test I can just say NavigateTo(AddUserPage.Path) rather than having to make a HomePage, click on User, and click on AddUser. If you compare Kevin's sample code from class to my code over in groupPageTests you'll see what I mean.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,224 @@
...
+
+            if (linkName == "Secret Santa")
+            {
+                Assert.IsTrue(Driver.Url == HomePage.Path);
+            }
+
+            else
+            {
+                Assert.IsTrue(Driver.Url.EndsWith(linkName));
+            }
```
BHoward85: Good use of the DataRow attribute

SecretSanta/test/SecretSanta.Web.UITests/DriverPages/AddGroupsPage.cs
BHoward85: I see, its a good way of navigation.

SecretSanta/test/SecretSanta.Web.UITests/DriverPages/HomePage.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using OpenQA.Selenium;
+
+namespace SecretSanta.Web.UITests.DriverPages
+{
+    public class HomePage
+    {
+        public const string Path = "https://localhost:44325/";
```
Keboo: I do like how you setup the nested path/slugs for navigation.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,224 @@
...
+
+            var editPage = new EditUsersPage(Driver);
+
+            Assert.IsTrue(Driver.Url.EndsWith(EditUsersPage.Slug + "/" + editPage.GetUserId));
+        }
+
+        [DataRow("Secret Santa")]
+        [DataRow("Groups")]
+        [TestMethod]
+        public void CanNavigateFromUsersPageToNavBarPages(string linkName)
```
Keboo: 👍

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,224 @@
...
+        [DataRow("Users")]
+        [TestMethod]
+        public void CanNavigateFromEditUsersPageToNavBarPages(string linkName)
+        {
+            Driver.Navigate().GoToUrl(EditUsersPage.Path);
+            var page = new EditUsersPage(Driver);
+            var navbarLink = page.GetNavbarLink(linkName);
+            navbarLink.Click();
+
+            if (linkName == "Secret Santa")
```
Keboo: In general we try to avoid having conditions in our unit tests based on passed in data. A better option is to have this method take a second parameter such as the URL and do the Assert based on that.

SecretSanta/test/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,224 @@
...
+            {
+                Assert.IsTrue(Driver.Url == HomePage.Path);
```
Keboo: Typically you want to do an `Assert.AreEqual` rather than an `Assert.IsTrue` with an `==`. Simply because on a failure you will get a more useful error message.

PR #156, opened 2019-03-13
Assignment9

COsborn2:


SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,203 @@
...
+                UsersPage usersPage = new UsersPage(Driver);
+                IWebElement homeLink = usersPage.HomeLink;
+
+                homeLink.Click();
+
+                Assert.AreEqual(RootUrl, Driver.Url);
+            });
+        }
+
+        private UsersPage CreateUser(IWebDriver Driver, string firstName, string lastName)
```
raymondshiner: Since CreateUser is private, I don't see why you would need to pass in Driver. I'd change this to just be (string fname, string lname) and just use the class level like you do in all of your other tests.

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,203 @@
...
+                var rootUri = new Uri(RootUrl);
+                Driver.Navigate().GoToUrl(new Uri(rootUri, UsersPage.Slug));
+                var page = new UsersPage(Driver);
+
+                page.AddUser.Click();
+
+                Assert.IsTrue(Driver.Url.EndsWith(AddUsersPage.Slug));
+            });
+        }
+
```
raymondshiner: I think you should have a CanNavigateToEditPage here

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
COsborn2: Maybe Im misinterpreting this, but I don't have a class level driver. My UITester.UITest takes an anonymous method with a parameter of IWebDriver, which I am passing in when my action is invoked. Hopefully that clears things up a little bit...

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
COsborn2: Good call, thanks.

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
raymondshiner: My apologies, I didn't realize that you didn't have a class level driver. I didn't fully understand how the UITester was set up. Thanks for clearing up my confusion!

SecretSanta/test/SecretSanta.Web.UITests/EditUsersPage.cs
```diff
@@ -0,0 +1,25 @@
+﻿using OpenQA.Selenium;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace SecretSanta.Web.UITests
+{
+    class EditUsersPage
```
Keboo: In general you always want to specify protection modifiers.

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,203 @@
...
+            var addUserPage = new AddUsersPage(Driver);
+
+            addUserPage.UserFirstNameTextBox.SendKeys(firstName);
+            addUserPage.UserLastNameTextBox.SendKeys(lastName);
+            addUserPage.SubmitButton.Click();
+            return page;
+        }
+    }
+
+    public class UITester
```
Keboo: This class should really be in its own file

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,203 @@
...
+            foreach (IWebDriver webDriver in Drivers)
+            {
+                try
+                {
+                    action(webDriver);
+                }
+                catch (Exception ex)
+                {
+                    string fileName = $"{action.Method.Name}-{webDriver.GetType()}.png";
+                    fileName = string.Join("", fileName.Split(Path.GetInvalidFileNameChars()));
```
Keboo: Very clever way of doing a `string.Replace` with a collection. :+1:

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,203 @@
...
+                    string currentDirectory = Directory.GetParent(Environment.CurrentDirectory).Parent.Parent.FullName;
```
Keboo: This is making an assumption on the current working directory. In addition this could fail if the working directory doesn't have two parents.

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,203 @@
...
+                    string path = Path.Combine(currentDirectory, FolderName, fileName);
+
+                    Screenshot screenshot = ((ITakesScreenshot)webDriver).GetScreenshot();
```
Keboo: This works because the current driver supports screenshots, but what happens when the Driver doesn't implement ITakesScreenshot? Better to type check rather than simply cast.
```suggestion
                if (Driver is ITakesScreenshot takesScreenshot)
                {
                    var screenshot = takesScrrenshot.GetScreenshot();
```

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
Keboo: This is a very interesting class. It is a clever way to try and handle multiple `IWebDrivers`. 
However this functionality is all stuff that is already built into the unit test framework. A more common approach is to have the unit tests in some common base class. Then in derived classes initialize whatever Driver in the test initialize method.

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,203 @@
...
+    {
+        private const string RootUrl = "https://localhost:44308/";
+        public TestContext TestContext { get; set; }
+
+        public UITester Tester { get; set; }
+
+        [TestInitialize]
+        public void Init()
+        {
+            IWebDriver ChromeWebDriver = new ChromeDriver(Path.GetFullPath("."));
```
Keboo: Minor thing. This should be `chromeWebDriver`

PR #158, opened 2019-03-13
Assignment 9 - Bryan Caesar

bcaesar42:


SecretSanta/test/SecretSanta.Web.UITests/GroupPageTests.cs
```diff
@@ -0,0 +1,184 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using OpenQA.Selenium;
+using OpenQA.Selenium.Chrome;
+
+namespace SecretSanta.Web.UITests
+{
+    [TestClass]
+    public class GroupPageTests
+    {
+        private const string RootUrl = "https://localhost:44331/";
```
Chester-Southwood: Early on submission, but it seems like you were able to connect it with your computer's port number.

SecretSanta/test/SecretSanta.Api.Tests/Controllers/GiftControllerTests.cs
```diff
@@ -22,6 +22,18 @@ public static void ConfigureAutoMapper(TestContext context)
             Mapper.Initialize(cfg => cfg.AddProfile(new AutoMapperProfileConfiguration()));
         }
 
+        [ClassInitialize]
```
Chester-Southwood: I'm not sure why you made the setup and cleanup for these files since we're not required to alter any of the given files from Kevin's startup. It's not a bad idea, but currently I only see the skeleton code due to this being a early submission.

SecretSanta/test/SecretSanta.Web.UITests/Pages/UsersPage.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            return deleteLinks.Single(x => x.GetAttribute("onclick")
+                              .EndsWith($"{userFirstName} {userLastName}?')"));
+        }
+
+        public IWebElement GetEditLink(string userFirstName, string userLastName)
+        {
+            string userId = GetUserId(userFirstName, userLastName);
+
+            return Driver.FindElement(By.CssSelector($"ul>li>a[href^=\"/Users/Edit/{userId}\"].button"));
```
Keboo: Nice selector :+1:

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,201 @@
...
+        [TestInitialize]
+        public void Init()
+        {
+            Driver = new ChromeDriver(Path.GetFullPath("."));
+        }
+
+        [TestCleanup]
+        public void Cleanup()
+        {
+            Driver?.Quit();
```
Keboo: In this case I don't think you want the null propagation operator here. If the Driver is ever null, something has gone wrong and probably needs to be fixed.

SecretSanta/test/SecretSanta.Web.UITests/UserPageTests.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            //Act
+            page.GetEditLink(userFirstName, userLastName).Click();
+
+            //Assert
+            //Assert.AreEqual<string>($"https://localhost:44302/{EditUsersPage.Slug}/{createdUserId}", Driver.Url);
+            Assert.IsTrue(Driver.Url.EndsWith($"{EditUsersPage.Slug}/{createdUserId}"));
+
+            // Cleanup
+            DeleteUser(userFirstName, userLastName);
```
Keboo: There is a bit of a problem with doing cleanup inside of the unit test. If an `Assert` call fails for some reason, it will throw an exception, so yo will never reach this cleanup. Usually it is best to do this kind of work in the test cleanup method.

PR #160, opened 2019-03-14
Assignment9 - Christopher Sturm

Sturminator:


SecretSanta/SecretSanta.Web.UITests/Pages/UsersPage.cs
```diff
@@ -0,0 +1,57 @@
...
+
+                        return text;
+                    })
+                    .ToList();
+            }
+        }
+
+        public IWebElement GetEditLink(string username)
+        {
+            ReadOnlyCollection<IWebElement> userElements =
```
ConnerV42: Very readable GetEditLink method, I like the way you did yours better than the way I coded mine.

SecretSanta/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,147 @@
...
+            deleteLink.Click();
+
+            Driver.SwitchTo().Alert().Accept(); // handle alert      
+            List<string> groupNames = up.UserList;
+            Assert.IsFalse(groupNames.Contains(fullName));
+        }
+
+        private UsersPage CreateUser(string rootUrl, string firstName, string lastName)
+        {
+            Driver.Navigate().GoToUrl(new Uri(new Uri(rootUrl), UsersPage.Slug));
```
ConnerV42: I don't think you need to instantiate a Uri inside of a Uri instantiation

SecretSanta/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            IWebElement deleteLink = up.GetDeleteLink(fullName);
+            deleteLink.Click();
+
+            Driver.SwitchTo().Alert().Accept(); // handle alert      
+            List<string> groupNames = up.UserList;
+            Assert.IsFalse(groupNames.Contains(fullName));
+        }
+
+        private UsersPage CreateUser(string rootUrl, string firstName, string lastName)
```
ConnerV42: You don't need to be passing the rootUrl here, because it is a class level constant property that you can access from anywhere.

SecretSanta/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,147 @@
...
+            editLink.Click();
+
+            Assert.AreEqual<string>(firstName, editPage.FirstNameTextBox.GetAttribute("value"));
+            Assert.AreEqual<string>(lastName, editPage.LastNameTextBox.GetAttribute("value"));
+        }
+
+        [TestMethod]
+        public void CanAddUser()
+        {
+            string firstName = "First Name" + Guid.NewGuid().ToString("N");
```
grantdwoods: I see these three lines of code often. I would suggest making them into a method!

SecretSanta/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        [TestMethod]
+        public void CanEditUser()
+        {
+            string firstName = "First Name" + Guid.NewGuid().ToString("N");
+            string lastName = "Last Name" + Guid.NewGuid().ToString("N");
+            string fullName = firstName + " " + lastName;
+
+            UsersPage up = CreateUser(RootUrl, firstName, lastName);
```
grantdwoods: I would suggest not using thing names like "up" and "eup". While I can understand the context if I go hunting for the initialization, the code would be more readable if I didn't have to double check.

Also the coding guidelines suggest this: 
"DO NOT use abbreviations or contractions within identifier names."

SecretSanta/SecretSanta.Web.UITests/UsersPageTests.cs
grantdwoods: Agreed.

SecretSanta/SecretSanta.Web.UITests/Pages/EditUserPage.cs
```diff
@@ -0,0 +1,23 @@
+﻿using OpenQA.Selenium;
+using System;
+using System.Linq;
+
+namespace SecretSanta.Web.UITests.Pages
+{
+    public class EditUserPage
+    {
+        IWebDriver Driver { get; }
```
Keboo: You typically want to specify all protection modifiers.

SecretSanta/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace SecretSanta.Web.UITests
+{
+    [TestClass]
+    public class UsersPageTests
+    {
+        protected IWebDriver Driver { get; set; }
```
Keboo: This should be `private` unless a derived class needs to access it.

SecretSanta/SecretSanta.Web.UITests/UsersPageTests.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            string firstNameEdit = "First Name" + Guid.NewGuid().ToString("N");
+            string lastNameEdit = "Last Name" + Guid.NewGuid().ToString("N");
+            string fullNameEdit = firstNameEdit + " " + lastNameEdit;
+
+            editUserPage.FirstNameTextBox.SendKeys(firstNameEdit);
+            editUserPage.LastNameTextBox.SendKeys(lastNameEdit);
+            editUserPage.SubmitButton.Click();
+
+            List<string> users = usersPage.UserList as List<string>;
```
Keboo: If you use the `as` keyword to cast something you must check it for `null`. If the cast is always expected to succeed then use a direct cast instead.
In this case `usersPage.UserList` is already a list so the cast is not needed.

